PR #12, opened 2021-12-27
[Fix] Fix input names

SemyonBevzuk:
## Motivation
It was found that if you change the input name `input` in the config (`onnx_config.input_names`) to, for example, `image`, then the deployment does not work quite correctly.
This PR fixes this problem for OpenVINO and ONNXRuntime.

## Modification
- `mmdeploy/apis/openvino/utils.py`: Added replacement of the default input name with the value from `onnx_config`.
- `mmdeploy/codebase/mmdet/deploy/object_detection_model.py`: Added the ability to take the input name from `onnx_config`.
- `mmdeploy/utils/config_utils.py`: Added check for input name in `onnx_config`.
- `tests/test_apis/test_onnx2openvino.py`: Added test case with `onnx_config` to `test_get_input_info_from_cfg`.
- `tests/test_utils/test_util.py`: Added tests for function `is_dynamic_shape` with `onnx_config` in config.


lvhan028:
@AllentDan Please double check if it is consistent with SDK Model's spec

AllentDan:
This is a bug for all codebases. May update to codebases besides MMDet

AllentDan:
> @AllentDan Please double check if it is consistent with SDK Model's spec

SDK Models also have the above errors and need update codes too

mmdeploy/utils/config_utils.py
```diff
@@ -154,7 +154,7 @@ def is_dynamic_batch(deploy_cfg: Union[str, mmcv.Config],
 
 
 def is_dynamic_shape(deploy_cfg: Union[str, mmcv.Config],
-                     input_name: str = 'input') -> bool:
+                     input_name: str = None) -> bool:
```
AllentDan, line 157: the type hint could be Optional[str]

tests/test_apis/test_onnx2openvino.py
```diff
@@ -121,3 +121,15 @@ def test_get_input_info_from_cfg():
     input_info = get_input_info_from_cfg(deploy_cfg)
     assert input_info == expected_input_info, \
         'The expected value of \'input_info\' does not match the received one.'
+
+    # The case where the input name in 'onnx_config'
+    # is different from 'backend_config'.
+    onnx_config_input_name = 'image'
```
AllentDan, line 127: could the input name for testing be randomly generated?

mmdeploy/apis/openvino/utils.py
```diff
@@ -19,4 +41,9 @@ def get_input_info_from_cfg(deploy_cfg: mmcv.Config) -> Dict[str, List]:
     # The partition is not supported now. Set the id of model to 0.
     model_inputs = get_model_inputs(deploy_cfg)[0]
     input_info = model_inputs['opt_shapes']
+    onnx_config = deploy_cfg.get('onnx_config', None)
+    if onnx_config:
```
SingleZombie, line 45: It seems that these codes make the priority of `input_names` in `onnx_config` greater than that in `backend_config`. If the `input_names` is set in `backend_config` but it differs from that in `onnx_config`, then the names will be replaced by the names in `onnx_config`.

However, generally, we assume that the `onnx_config` is for ONNX export and `backend_config` is for backend operations. Users should ensure that the names are identical by themselves. Is it necessary to make these changes?

AllentDan:
Hi, @SemyonBevzuk. I've made some changes to the codes, could you please open the permission of your fork to me?

SemyonBevzuk:
Of course, access is open.
I will look at the rest of the comments today, but a little later.

mmdeploy/apis/openvino/utils.py
SemyonBevzuk: Yes, in the current version `input_names` from `onnx_config` has a higher priority.

It seems unclear to me that changing the input name in the config requires making changes in three places:
- `onnx_config.input_names`
- `onnx_config.dynamic_axes`
- `backend_config.model_inputs.opt_shapes`

By the way, it seems to me that there is no need to use a specific input name in `backend_config`. We could just write the default shape value to the config:
```
backend_config = dict(
    model_inputs = dict(
        default_shape = [1, 3, 800, 1344]))
```

mmdeploy/utils/config_utils.py
SemyonBevzuk: Done.

tests/test_apis/test_onnx2openvino.py
SemyonBevzuk: Added random name generation in tests `test_onnx2openvino` and `test_get_input_info_from_cfg`.

mmdeploy/apis/openvino/utils.py
grimoire: Sure.  we decide to alias `onnx_config` to `ir_config` and read `input_names` from it (guess we might support more IR in the future).
May I commit in this branch?

SemyonBevzuk: Yes, please.

grimoire: Hi, It seems that `onnx_config.input_names` and `onnx_config.dynamic_axes` might not share the same names.

For example, if I have a model with two inputs, one needs dynamic shapes to support and one does not need (such as binary-ops between scalar and tensor). Names in `input_names` and `dynamic_axes` would be different since we just need to set the axes of the dynamic input.

We have not begun to support models with multiple inputs yet. I am not sure if it is a good idea to require dynamic axes for all inputs even when the input does not need it. And I am not sure if there would be a similar problem on the inference engine side.

SemyonBevzuk: Okay, thanks for the clarification. Agree with you about dynamic axes.

At the moment (this may change in the near future) the Model Optimizer from OpenVINO requires shape for all inputs. So, I can add a default input shape to `backend_config.model_inputs` and the ability to manually specify the shapes for the input names. It might look like this:
```
backend_config = dict(
    model_inputs = dict(
        default_shape = [...],
        custom_input_name = [...]
    )
)
```

grimoire:
Now `dynamic_axes` and `opt_shape` accept both dictionary and list. @AllentDan please check if model can be dumped to sdk.

mmdeploy/apis/pytorch2onnx.py
```diff
@@ -28,6 +28,13 @@ def torch2onnx_impl(model: torch.nn.Module, input: torch.Tensor,
     backend = get_backend(deploy_cfg).value
     opset_version = pytorch2onnx_cfg.get('opset_version', 11)
 
+    input_names = pytorch2onnx_cfg['input_names']
+    output_names = pytorch2onnx_cfg['output_names']
+    dynamic_axes = pytorch2onnx_cfg.get('dynamic_axes', None)
+
+    if dynamic_axes is not None and not isinstance(dynamic_axes, Dict):
+        dynamic_axes = dict(zip(input_names + output_names, dynamic_axes))
```
SingleZombie, line 36: I think this line is important and our test cases should cover it.

AllentDan:
> both

LGTM on SDK side

mmdeploy/apis/pytorch2onnx.py
```diff
@@ -28,6 +28,13 @@ def torch2onnx_impl(model: torch.nn.Module, input: torch.Tensor,
     backend = get_backend(deploy_cfg).value
     opset_version = pytorch2onnx_cfg.get('opset_version', 11)
```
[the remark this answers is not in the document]
SemyonBevzuk: Done.

mmdeploy/codebase/mmcls/deploy/classification_model.py
```diff
@@ -32,16 +32,16 @@ def __init__(
...
         self.deploy_cfg = deploy_cfg
         self._init_wrapper(
             backend=backend, backend_files=backend_files, device=device)
 
     def _init_wrapper(self, backend: Backend, backend_files: Sequence[str],
                       device: str):
-        onnx_config = get_onnx_config(self.deploy_cfg)
-        output_names = onnx_config['output_names']
+        ir_config = get_ir_config(self.deploy_cfg)
+        output_names = ir_config['output_names']
```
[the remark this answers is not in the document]
SemyonBevzuk: Yes, we can. Done.

mmdeploy/utils/config_utils.py
```diff
@@ -154,7 +154,7 @@ def is_dynamic_batch(deploy_cfg: Union[str, mmcv.Config],
 
 
 def is_dynamic_shape(deploy_cfg: Union[str, mmcv.Config],
-                     input_name: str = 'input') -> bool:
+                     input_name: Optional[str] = None) -> bool:
```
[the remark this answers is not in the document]
SemyonBevzuk: Done.

mmdeploy/apis/pytorch2onnx.py
SemyonBevzuk: Test `test_torch2onnx` has been rewritten to check this case.

AllentDan (approved):
LGTM

SingleZombie (approved):
LGTM

PR #26, opened 2021-12-29
modified for correctly using ort on GPU

Stephenfang51:
The previous code works wrong, no matter you select which device, for example you select device 3
when call `ort.InferenceSession` functions, it just load model to device:0, after that you give a provider and `device_Id:3`, then load same model to device 3 (load twice)
so I simplified the process which works fine

### DIFF
--- a/mmdeploy/backend/onnxruntime/wrapper.py
+++ b/mmdeploy/backend/onnxruntime/wrapper.py
@@ -50,18 +50,12 @@ class ORTWrapper(BaseWrapper):
         else:
             logging.warning(f'The library of onnxruntime custom ops does \
             not exist: {ort_custom_op_path}')
-
-        sess = ort.InferenceSession(onnx_file, session_options)
-
+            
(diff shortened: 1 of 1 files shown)


AllentDan:
Hi, you can actually continue fixing this on the previous PR. Just pull first in the patch-1 branch, then add your codes, and finally commit push.

BTW, lint could be fixed through [this](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md#code-style)

Stephenfang51:
> Hi, you can actually continue fixing this on the previous PR. Just pull first in the patch-1 branch, then add your codes, and finally commit push.
> 
> BTW, lint could be fixed through [this](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md#code-style)

Thanks for reply.  already done with patch-1 branch.

AllentDan:
Thanks for your kindly contribution, close this pull request then.

PR #49, opened 2022-01-06
[Fix] Fix TensorRT 7 SSD

grimoire:
TensorRT 7 does not support dynamic clamp, which is used in `torch.nn.functional.normalize`.


RunningLeon (approved):
LGTM

SingleZombie (approved):
LGTM

PR #58, opened 2022-01-10
[Fix] Fixed device_id in tools/test.py for the CPU.

SemyonBevzuk:
## Motivation
It was found that `tools/test.py` with argument `--device cpu` crashes with the following error:
```bash
2022-01-11 01:45:25,318 - mmdeploy - INFO - Successfully loaded onnxruntime custom ops from             /home/sbevzyuk/mmdeployment_dir/MMDeploy/build/lib/libmmdeploy_onnxruntime_ops.so
Traceback (most recent call last):
  File "/home/sbevzyuk/mmdeployment_dir/MMDeploy/tools/test.py", line 141, in <module>
    main()
  File "/home/sbevzyuk/mmdeployment_dir/MMDeploy/tools/test.py", line 112, in main
    model = MMDataParallel(model, device_ids=[device_id])
  File "/home/sbevzyuk/anaconda3/envs/mmdeployment/lib/python3.7/site-packages/mmcv/parallel/data_parallel.py", line 35, in __init__
    super(MMDataParallel, self).__init__(*args, dim=dim, **kwargs)
  File "/home/sbevzyuk/anaconda3/envs/mmdeployment/lib/python3.7/site-packages/torch/nn/parallel/data_parallel.py", line 140, in __init__
    self.src_device_obj = torch.device(device_type, self.device_ids[0])
RuntimeError: Device index must not be negative`
```
This PR proposes a fix for this problem. 

## Modification
- `mmdeploy/backend/openvino/wrapper.py`: Fixed work with the dictionary for Python 3.8.
- `mmdeploy/utils/device.py`: The documentation of the `parse_device_id` function was updated.
- `tests/test_utils/test_util.py`: Added tests for function `parse_device_id`.
- `tools/test.py`: Added fix for `cpu`.

### DIFF
--- a/mmdeploy/utils/device.py
+++ b/mmdeploy/utils/device.py
@@ -1,24 +1,27 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+from typing import Optional
+
 import torch
 
 
-def parse_device_id(device: str) -> int:
-    """Parse cuda device index from a string.
+def parse_device_id(device: str) -> Optional[int]:
+    """Parse device index from a string.
 
     Args:
-        device (str): The typical style of string specifying cuda device,
-            e.g.: 'cuda:0'.
+        device (str): The typical style of string specifying device,
+            e.g.: 'cuda:0', 'cpu'.
 
     Returns:
-        int: The parsed device id, defaults to `0`.
-            If device is 'cpu' it will return `-1`.
+        Optional[int]: The return value depends on the type of device.
+            If device is 'cuda': cuda device index, defaults to `0`.
+            If device is 'cpu': `-1`.
+            Otherwise, `None` will be returned.
     """
     if device == 'cpu':
         return -1
-    device_id = 0
-    if len(device) >= 6:
-        device_id = torch.device(device).index
-    return device_id
+    if 'cuda' in device:
+        return parse_cuda_device_id(device)
+    return None
 
 
 def parse_cuda_device_id(device: str) -> int:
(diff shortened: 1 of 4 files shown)


tools/test.py
```diff
@@ -104,7 +104,10 @@ def main():
     # load the model of the backend
     model = task_processor.init_backend_model(args.model)
 
-    device_id = parse_device_id(args.device)
+    if args.device == 'cpu':
+        device_id = 0
+    else:
+        device_id = parse_device_id(args.device)
```
AllentDan, line 110: Seems `device_id` get set to 0 for both `cpu` and `cuda:0` device. Maybe we have pass None to device_ids in MMDataParallel like described [here](https://github.com/open-mmlab/mmcv/blob/b57825f3f1f38cf6033e0d6d36af644cc57465fa/mmcv/parallel/data_parallel.py#L28). Besides, `with_sync` also need to be taken into account.

mmdeploy/utils/device.py
```diff
@@ -11,6 +11,7 @@ def parse_device_id(device: str) -> int:
 
     Returns:
         int: The parsed device id, defaults to `0`.
+            If device is 'cpu' it will return `-1`.
     """
     if device == 'cpu':
```
AllentDan, line 16: The docstring may get updated. `parse_device_id` parse both cpu and gpu devices

tools/test.py
SemyonBevzuk: Thanks, fixed.

mmdeploy/utils/device.py
SemyonBevzuk: Updated the documentation and implementation a bit.

AllentDan (approved):
LGTM

grimoire (approved):
LGTM

RunningLeon (approved):
LGTM

PR #61, opened 2022-01-11
[Enhancement]: Remove deepcopy in patch_model

SingleZombie:
## Motivation

When `patch_model`, we have to deepcopy a model, which increases the memory overhead.

## Modification

Remove `deepcopy` and add comments.


grimoire (approved):

RunningLeon (approved):
LGTM

Issue #79, opened 2022-01-15
MMOCR example from SDK


pk-cb:
Steps completed:
1. SDK Build
2. Example build

While running the ocr demo in `MMDeploy/build/install/example/` I'm getting this error:
```
./ocr: error while loading shared libraries: libmmdeploy_core.so: cannot open shared object file: No such file or directory
```

aliml92:
same here. Would've been great if the problem solved

AllentDan:
Hi, what if we append `-Wl,--disable-new-dtags` to line 23 of  `cmake/MMDeployConfig.cmake.in`?

tehkillerbee:
I don't have this issue, provided I set LD_LIBRARY_PATH according to the documentation.
```
export LD_LIBRARY_PATH=/home/${USER}/MMDeploy/build/install/lib
```

How did you build your SDK?

lvhan028:
fixed in #80

PR #80, opened 2022-01-17
[Fix] fix missing deploy_core

AllentDan:
#79

### DIFF
--- a/demo/csrc/CMakeLists.txt
+++ b/demo/csrc/CMakeLists.txt
@@ -5,6 +5,11 @@ project(mmdeploy-example)
 find_package(OpenCV REQUIRED)
 find_package(MMDeploy REQUIRED)
 
+set(MMDeploy_LIBS ${MMDEPLOY_LIBS}
+    -Wl,--no-as-needed ${MMDEPLOY_DYNAMIC_MODULES} -Wl,--as-needed
+    -Wl,--whole-archive ${MMDEPLOY_STATIC_MODULES} -Wl,--no-whole-archive
+    -Wl,--disable-new-dtags)
+
 function(add_example name)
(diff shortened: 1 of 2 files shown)


lvhan028:
I think it's better for us to dig more about `rpath` and `runpath`. 
From https://www.qt.io/blog/2011/10/28/rpath-and-runpath, it seems '-Wl,--disable-new-dtags' is not a recommended method.

AllentDan:
> I think it's better for us to dig more about `rpath` and `runpath`. From https://www.qt.io/blog/2011/10/28/rpath-and-runpath, it seems '-Wl,--disable-new-dtags' is not a recommended method.

Maybe we can remind users to add our lib to LD_LIBRARY_PATH

lvhan028:
> > I think it's better for us to dig more about `rpath` and `runpath`. From https://www.qt.io/blog/2011/10/28/rpath-and-runpath, it seems '-Wl,--disable-new-dtags' is not a recommended method.
> 
> Maybe we can remind users to add our lib to LD_LIBRARY_PATH

Agreed

demo/csrc/CMakeLists.txt
```diff
@@ -5,6 +5,11 @@ project(mmdeploy-example)
 find_package(OpenCV REQUIRED)
 find_package(MMDeploy REQUIRED)
 
+set(MMDeploy_LIBS ${MMDEPLOY_LIBS}
```
lvhan028, line 8: How about ```target_link_libraries(${name} ${MMDeploy_LIBS} -Wl,--disable-new-dtags ...)```?

lvhan028 (approved):
LGTM

lzhangzz (approved):

PR #93, opened 2022-01-20
Replace convert_syncbatchnorm in mmseg

RunningLeon:
## Motivation

Remove `convert_syncbatchnorm` and use `mmcv.cnn.utils.revert_sync_batchnorm` since `revert_sync_batchnorm` is supported in mmcv by PR https://github.com/open-mmlab/mmcv/pull/1253. 

## Modification

- [x] Replace `convert_syncbatchnorm` with `revert_sync_batchnorm`
- [x] Remove `mmdeploy.codebase.mmseg.deploy.utils.convert_syncbatchnorm` and its unit test.
- [x] Update docs that [UNet](https://github.com/open-mmlab/mmsegmentation/blob/master/configs/unet/fcn_unet_s5-d16_4x4_512x1024_160k_cityscapes.py) in mmseg only supports static shape due to concat failure in [UpConvBlock](https://github.com/open-mmlab/mmsegmentation/blob/7512f05990eb66bba3653cb4d5f478965bf41bd7/mmseg/models/utils/up_conv_block.py#L99)

## BC-breaking (Optional)

None


mmdeploy/codebase/mmseg/deploy/segmentation.py
```diff
@@ -225,8 +225,7 @@ def evaluate_outputs(model_cfg,
                 to `False`.
         """
         if out:
-            logger = get_root_logger()
-            logger.info(f'\nwriting results to {out}')
+            logging.info(f'\nwriting results to {out}')
```
grimoire, line 228: use logger in mmcv instead of logging.

AllentDan:
Shall we rewrite the upsampling module to rescale the input `x` to the size of `skip`. This is actually what I did to mmocr

AllentDan (approved):

AllentDan:
> Shall we rewrite the upsampling module to rescale the input `x` to the size of `skip`. This is actually what I did to mmocr

We can make `UNet` dynamically deployed in the future

RunningLeon:
> Shall we rewrite the upsampling module to rescale the input `x` to the size of `skip`. This is actually what I did to mmocr
@AllentDan 
`Upsample` is set through config and there are many cases for it. Maybe should consider it in another pr if there has a request.

RunningLeon:
Dynamic shape support with UNet will be done in mmseg repo. Related issue https://github.com/open-mmlab/mmsegmentation/issues/1236

grimoire (approved):

PR #99, opened 2022-01-23
[Enhancement] Simplify non batch nms

grimoire:
## Motivation

Batched nms require a lot of index node to support batch inference, which is not necessary.

## Modification

Check if the deploy_cfg require static shape and batch size is 1.
Simplify nms.


RunningLeon (approved):
LGTM

RunningLeon:
@SingleZombie Hi, any other comment on this PR?

SingleZombie (approved):
LGTM

PR #101, opened 2022-01-24
[Fix]fix check_env

VVsssssk:
## Motivation

Fix check_env.py script to check codebase version.

## Modification

Fix check_env.py

## Use cases (Optional)

python tools/check_env.py

### DIFF
--- a/tools/check_env.py
+++ b/tools/check_env.py
@@ -1,4 +1,5 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+
 from mmcv.utils import collect_env as collect_base_env
 from mmcv.utils import get_git_hash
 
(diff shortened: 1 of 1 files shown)


tools/check_env.py
```diff
@@ -54,8 +55,81 @@ def check_backend():
...
+
+
+def check_codebase():
+    ops = {
+        '>=': (lambda x, y: x >= y),
+        '==': (lambda x, y: x == y),
+        '<=': (lambda x, y: x <= y),
+        '': (lambda x, y: True)
+    }
+    with open('../requirements/optional.txt', 'r') as file:
```
RunningLeon, line 73: file path may not exist if calling on different directory.

tools/check_env.py
```diff
@@ -54,8 +55,81 @@ def check_backend():
     logger.info(f'openvino_is_avaliable: {openvino_apis.is_available()}')
 
 
+def get_version(s):
+    return ''.join([c for c in s if c.isdigit() or c == '.'])
+
+
+def get_op(s):
+    return ''.join([c for c in s if c in '><='])
```
grimoire, line 63: Use Regular Expression instead of parse string.

tools/check_env.py
```diff
@@ -54,8 +55,81 @@ def check_backend():
...
+        '': (lambda x, y: True)
+    }
+    with open('../requirements/optional.txt', 'r') as file:
+        try:
+            import mmcls
+        except ImportError:
+            mmcls_version = None
+        else:
+            mmcls_version = mmcls.__version__
+        requirement = file.readline()
```
grimoire, line 80: Not safe to check packages one by one.
The package order in `optional.txt` might change in the future.

RunningLeon: Version requirements may change according to [PEP ](https://www.python.org/dev/peps/pep-0440/#version-specifiers) and this check may fail at some cases.

RunningLeon (approved):
LGTM

grimoire (approved):

PR #133, opened 2022-02-07
Added TensorRT version check

tehkillerbee:
On TensorRT version < 7.2,  `setOptimizationProfileAsync(...)` is not available. Instead, `context_->setOptimizationProfile(0);`  can be used.

This PR adds a TensorRT version check to ensure mmdeploy builds on older versions of TensorRT where  setOptimizationProfileAsync(...) is not available.


azuryl:
it still occoured in https://github.com/open-mmlab/mmdeploy/issues/2260

PR #141, opened 2022-02-09
[Enhancement] upgrade isort in pre-commit config

lvhan028:
## Motivation
https://github.com/open-mmlab/mmflow/pull/87

## Modification

1. Remove seed-isort-config in pre-commit hook
2. Upgrade isort to v5.10.1 (latest version)
3. Replace known_standard_library with extra_standard_library


mmdeploy/backend/ncnn/wrapper.py
```diff
@@ -8,6 +8,7 @@
 
 from mmdeploy.utils import Backend, get_root_logger
 from mmdeploy.utils.timer import TimeCounter
+
```
AllentDan, line 11: would this change make mmdeploy a third extra library?

RunningLeon (approved):
LGTM

mmdeploy/backend/ncnn/wrapper.py
lvhan028: You mean `extra_standard_library` making mmdeploy a third extra library?

MeowZheng (approved):

mmdeploy/backend/ncnn/wrapper.py
AllentDan: I was meaning that the previous codes made `mmdeploy` a third-party library. Now it gets fixed then

AllentDan (approved):
LGTM

PR #167, opened 2022-02-17
[Enhancement] Optimize multilevel roi align

grimoire:
Use `fma` to perform interpolate.

Test input:

```
rois shape: torch.Size([1000, 5])
feat 0 shape: torch.Size([1, 256, 200, 304])
feat 1 shape: torch.Size([1, 256, 100, 152])
feat 2 shape: torch.Size([1, 256, 50, 76])
feat 3 shape: torch.Size([1, 256, 25, 38])
```

Note that there is not much speed up( less than 0.5ms). Since load feat (which is randomly accessed, can not be cached) would cost a lot of time. I have no idea how to perform the optimize.

Or should I add a fp16 implementation?

### DIFF
--- a/csrc/backend_ops/tensorrt/multi_level_roi_align/trt_multi_level_roi_align_kernel.cu
+++ b/csrc/backend_ops/tensorrt/multi_level_roi_align/trt_multi_level_roi_align_kernel.cu
@@ -1,4 +1,5 @@
 // Copyright (c) OpenMMLab. All rights reserved.
+#include <float.h>
 #include <stdio.h>
 
 #include <algorithm>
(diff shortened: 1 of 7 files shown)


AllentDan (approved):
LGTM except that I find suddenly that roi align in MMCV supports `maxpooling` mode.

csrc/backend_ops/tensorrt/multi_level_roi_align/trt_multi_level_roi_align_kernel.cu
```diff
@@ -62,18 +58,21 @@ __device__ scalar_t roi_align_single(const scalar_t *bottom_data, const int roi_
   return output_val;
 }
 
-template <typename scalar_t>
-__global__ void roi_extractor_kernel(scalar_t *output, const scalar_t *bottom_rois,
-                                     FeatData feat_data, const int sample_num,
-                                     const float roi_scale_factor, const int finest_scale,
-                                     const int pooled_height, const int pooled_width,
-                                     const bool aligned, int nThreads) {
+template <typename scalar_t, bool aligned>
```
lvhan028, line 61: What's the necessity of adding a non-type template parameter `aligned`?

grimoire: I guess there would be less jump if aligned is constant. Which would benefit block level optimize.

AllentDan (approved):
LGTM

lvhan028 (approved):
LGTM

PR #168, opened 2022-02-18
[Enhancement] Update pad logic in detection heads

grimoire:
It is ugly to put an `if backend == 'tensorrt'` branch inside each head.
This PR add a rewrite to enable pad only when backend is tensorrt.


mmdeploy/codebase/mmdet/models/dense_heads/rpn_head.py
```diff
@@ -145,7 +144,7 @@ def rpn_head__get_bboxes(ctx,
 
 
 @FUNCTION_REWRITER.register_rewriter(
-    'mmdet.models.dense_heads.RPNHead.get_bboxes', backend='ncnn')
+    'mmdet.models.dense_heads.RPNHead.get_bboxes', backend=Backend.NCNN.value)
```
VVsssssk, line 147: Do we need to instead all func backend=str to backend=Backend.XXX.value?

tests/test_codebase/test_mmdet/test_mmdet_utils.py
```diff
@@ -29,6 +30,15 @@ def test_pad_with_value():
     assert np.allclose(padded_x.sum(), x.sum(), rtol=1e-03, atol=1e-05)
 
 
+def test_pad_with_value_if_necessary():
```
RunningLeon, line 33: should test with ort and trt since there are two rewrting functions

mmdeploy/codebase/mmdet/models/dense_heads/rpn_head.py
grimoire: We better do so, in case we might change the string in the future.
But not in this PR.

tests/test_codebase/test_mmdet/test_mmdet_utils.py
grimoire: I think this is not necessary since codecov has covered most cases.

RunningLeon:
could fix ci errors

RunningLeon (approved):
LGTM

VVsssssk (dismissed):

RunningLeon (approved):
LGTM

VVsssssk (approved):

PR #185, opened 2022-02-25
[Fix]use float in scale factor

AllentDan:
Fix a type error when computing scale_factor in rewriting interpolate.

### DIFF
--- a/mmdeploy/pytorch/functions/interpolate.py
+++ b/mmdeploy/pytorch/functions/interpolate.py
@@ -90,7 +90,8 @@ def interpolate__tensorrt(
                 'tensor. Which is not available for custom ops. Computed scale'
                 '_factor might be the right way to get final shape.')
             scale_factor = [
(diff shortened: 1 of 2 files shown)


mmdeploy/pytorch/functions/interpolate.py
```diff
@@ -90,7 +90,8 @@ def forward(g, input, scale_factor, align_corners):
                 'tensor. Which is not available for custom ops. Computed scale'
                 '_factor might be the right way to get final shape.')
             scale_factor = [
-                s_out / s_in for s_out, s_in in zip(size, input_size[2:])
+                float(s_out / s_in)
+                for s_out, s_in in zip(size, input_size[2:])
```
RunningLeon, line 94: should test unit be updated as well to cover this case?

grimoire (approved):

RunningLeon (approved):
LGTM

PR #208, opened 2022-03-04
Add DefaultFormatBundle

AllentDan:
## Motivation

To add DefaultFormatBundle to preprocess for SDK.

## Changed
**modified**
- csrc/preprocess/cpu/CMakeLists.txt
- csrc/preprocess/cuda/CMakeLists.txt
- mmdeploy/utils/export_info.py: remove the replacement from *DefaultFormatBundle* to *Image2Tensor*
**Added**
- csrc/preprocess/cuda/default_format_bundle_impl.cpp
- csrc/preprocess/cpu/default_format_bundle_impl.cpp
- csrc/preprocess/transform/default_format_bundle.cpp
- csrc/preprocess/transform/default_format_bundle.h
- tests/test_csrc/preprocess/test_default_format_bundle.cpp

## Note
I am not sure but I just dismissed `DC` part and label stuff for the **DefaultFormatBundle** in mmdet preprocess. Codes are below:
```python
    def __call__(self, results):
        """Call function to transform and format common fields in results.

        Args:
            results (dict): Result dict contains the data to convert.

        Returns:
            dict: The result dict contains the data that is formatted with \
                default bundle.
        """

        if 'img' in results:
            img = results['img']
            if self.img_to_float is True and img.dtype == np.uint8:
                # Normally, image is of uint8 type without normalization.
                # At this time, it needs to be forced to be converted to
                # flot32, otherwise the model training and inference
                # will be wrong. Only used for YOLOX currently .
                img = img.astype(np.float32)
            # add default meta keys
            results = self._add_default_meta_keys(results)
            if len(img.shape) < 3:
                img = np.expand_dims(img, -1)
            img = np.ascontiguousarray(img.transpose(2, 0, 1))
            results['img'] = DC(to_tensor(img), stack=True)
        for key in ['proposals', 'gt_bboxes', 'gt_bboxes_ignore', 'gt_labels']:
            if key not in results:
                continue
            results[key] = DC(to_tensor(results[key]))
        if 'gt_masks' in results:
            results['gt_masks'] = DC(results['gt_masks'], cpu_only=True)
        if 'gt_semantic_seg' in results:
            results['gt_semantic_seg'] = DC(
                to_tensor(results['gt_semantic_seg'][None, ...]), stack=True)
        return results

    def _add_default_meta_keys(self, results):
        """Add default meta keys.

        We set default meta keys including `pad_shape`, `scale_factor` and
        `img_norm_cfg` to avoid the case where no `Resize`, `Normalize` and
        `Pad` are implemented during the whole pipeline.

        Args:
            results (dict): Result dict contains the data to convert.

        Returns:
            results (dict): Updated result dict contains the data to convert.
        """
        img = results['img']
        results.setdefault('pad_shape', img.shape)
        results.setdefault('scale_factor', 1.0)
        num_channels = 1 if len(img.shape) < 3 else img.shape[2]
        results.setdefault(
            'img_norm_cfg',
            dict(
                mean=np.zeros(num_channels, dtype=np.float32),
                std=np.ones(num_channels, dtype=np.float32),
                to_rgb=False))
        return results
```

### DIFF
--- a/csrc/preprocess/transform/default_format_bundle.cpp
+++ b/csrc/preprocess/transform/default_format_bundle.cpp
@@ -1,0 +1,75 @@
+// Copyright (c) OpenMMLab. All rights reserved.
+
+#include "default_format_bundle.h"
+
+#include <cassert>
+
+#include "archive/json_archive.h"
+#include "core/tensor.h"
+
+namespace mmdeploy {
+
+DefaultFormatBundleImpl::DefaultFormatBundleImpl(const Value& args) : TransformImpl(args) {
+  if (args.contains("img_to_float") && args["img_to_float"].is_boolean()) {
+    arg_.img_to_float = args["img_to_float"].get<bool>();
+  }
+  if (args.contains("pad_val") && args["pad_val"].is_object()) {
+    for (auto& [key, _] : arg_.pad_val)
+      if (args["pad_val"].contains(key) && args["pad_val"][key].is_number_float()) {
+        arg_.pad_val[key] = args["pad_val"][key].get<float>();
+      }
+  }
+}
+
+Result<Value> DefaultFormatBundleImpl::Process(const Value& input) {
+  MMDEPLOY_DEBUG("DefaultFormatBundle input: {}", to_json(input).dump(2));
+  Value output = input;
+  if (input.contains("img")) {
+    Tensor tensor = input["img"].get<Tensor>();
+    OUTCOME_TRY(output["img"], ToFloat32(tensor, arg_.img_to_float));
+  }
+
+  Tensor tensor = output["img"].get<Tensor>();
+  // set default meta keys
+  for (auto v : tensor.shape()) {
+    output["pad_shape"].push_back(v);
+  }
+  output["scale_factor"].push_back(1.0);
+  int channel = tensor.shape()[3];
+  for (int i = 0; i < channel; i++) {
+    output["img_norm_cfg"]["mean"].push_back(0.0);
+    output["img_norm_cfg"]["std"].push_back(1.0);
+  }
+  output["img_norm_cfg"]["to_rgb"] = false;
+
+  // transpose
+  OUTCOME_TRY(output["img"], HWC2CHW(tensor));
+
+  MMDEPLOY_DEBUG("DefaultFormatBundle output: {}", to_json(output).dump(2));
+  return output;
+}
+
+DefaultFormatBundle::DefaultFormatBundle(const Value& args, int version) : Transform(args) {
+  auto impl_creator =
+      Registry<DefaultFormatBundleImpl>::Get().GetCreator(specified_platform_, version);
+  if (nullptr == impl_creator) {
+    MMDEPLOY_ERROR("'DefaultFormatBundle' is not supported on '{}' platform", specified_platform_);
+    throw std::domain_error("'DefaultFormatBundle' is not supported on specified platform");
+  }
+  impl_ = impl_creator->Create(args);
+}
+
+class DefaultFormatBundleCreator : public Creator<Transform> {
+ public:
+  const char* GetName() const override { return "DefaultFormatBundle"; }
+  int GetVersion() const override { return version_; }
+  ReturnType Create(const Value& args) override {
+    return std::make_unique<DefaultFormatBundle>(args, version_);
+  }
+
+ private:
+  int version_{1};
+};
+REGISTER_MODULE(Transform, DefaultFormatBundleCreator);
+MMDEPLOY_DEFINE_REGISTRY(DefaultFormatBundleImpl);
+}  // namespace mmdeploy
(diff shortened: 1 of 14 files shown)


csrc/preprocess/transform/default_format_bundle.h
```diff
@@ -0,0 +1,62 @@
...
+ *  - gt_labels: (1)to tensor, (2)to DataContainer
+ *  - gt_masks: (1)to tensor, (2)to DataContainer (cpu_only=True)
+ *  - gt_semantic_seg: (1)unsqueeze dim-0 (2)to tensor, \
+ *                     (3)to DataContainer (stack=True)
+ *
+ */
+class MMDEPLOY_API DefaultFormatBundleImpl : public TransformImpl {
+ public:
+  DefaultFormatBundleImpl(const Value& args);
+  ~DefaultFormatBundleImpl() = default;
```
lvhan028, line 28: miss `override`

lvhan028, line 28: miss `override`

csrc/preprocess/transform/default_format_bundle.h
```diff
@@ -0,0 +1,62 @@
...
+  using ArgType = struct default_format_bundle_arg_t;
+
+ protected:
+  ArgType arg_;
+};
+
+class MMDEPLOY_API DefaultFormatBundle : public Transform {
+ public:
+  explicit DefaultFormatBundle(const Value& args, int version = 0);
+  ~DefaultFormatBundle() = default;
```
lvhan028, line 50: miss `override`

lvhan028, line 50: miss `override`

csrc/preprocess/transform/default_format_bundle.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "archive/json_archive.h"
+#include "core/tensor.h"
+
+namespace mmdeploy {
+
+DefaultFormatBundleImpl::DefaultFormatBundleImpl(const Value& args) : TransformImpl(args) {
+  if (args.contains("img_to_float") && args["img_to_float"].is_boolean()) {
+    arg_.img_to_float = args["img_to_float"].get<bool>();
+  }
+  if (args.contains("pad_val") && args["pad_val"].is_object()) {
```
lvhan028, line 16: What `pad_val` is used for? I didn't find the usage in the `*impl` classes

csrc/preprocess/transform/default_format_bundle.cpp
```diff
@@ -0,0 +1,81 @@
...
+Result<Value> DefaultFormatBundleImpl::Process(const Value& input) {
+  MMDEPLOY_DEBUG("DefaultFormatBundle input: {}", to_json(input).dump(2));
+  Value output = input;
+  if (input.contains("img")) {
+    Tensor tensor = input["img"].get<Tensor>();
+    OUTCOME_TRY(output["img"], ToFloat32(tensor, arg_.img_to_float));
+  }
+
+  Tensor tensor = output["img"].get<Tensor>();
+  // set default meta keys
```
lvhan028, line 33: I don't think the default key should be set here. 

If `Resize` goes after `DefaultFormatBundle` in the preprocessing pipeline, the `scale_factor` generated by `DefaultFormatBundle` will shadow the if-else branch `arg_.img_scale`  of  `Resize`.

csrc/preprocess/transform/default_format_bundle.h
AllentDan: Shall we add `override` to all the constructor and destructor functions for the classes inherited from `Transform` and `TransformImpl`?

lvhan028: It's a clang-tidy warning, not an error. The rest ones can be fixed in another PR

lvhan028:
It failed to export the default value of `DefaultFormatBundle`.
Here is part of the dump info when exporting yolox model:
```json
{
    "type": "DefaultFormatBundle"
}
```
But the expected one is:
```json
{
    "type": "DefaultFormatBundle",
    "img_to_float": "true"
}
```

lvhan028:
> > It failed to export the default value of `DefaultFormatBundle`. Here is part of the dump info when exporting yolox model:
> > ```json
> > {
> >     "type": "DefaultFormatBundle"
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But the expected one is:
> > ```json
> > {
> >     "type": "DefaultFormatBundle",
> >     "img_to_float": "true"
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @AllentDan
> 
> Well, I just used `img_to_float = true` for the default setting in SDK, which the same with python setting.

Suppose codebase changes `img_to_float` to `false`, SDK has to change code accordingly. I don't think it's good for maintenance.

AllentDan:
As we implement other transformations like [resize](https://github.com/open-mmlab/mmdeploy/blob/df4e9e6caec3cfb95f6e474d76ba5e5b8cf370ea/csrc/preprocess/transform/resize.h#L25), we put many other default values there and don't dump them if users don't clarify it in configs. Besides, if codebases changed the default value of the function, I don't think they only changed the value while keeping other codes the same. Therefore, SDK still has to change accordingly.

lvhan028 (approved):
LGTM

lzhangzz (approved):
LGTM

PR #243, opened 2022-03-16
[Docs] add faq to `How to install mmdeploy on Jetsons`

AllentDan:
Add a FAQ to `How to install mmdeploy on Jetsons`

### DIFF
--- a/csrc/backend_ops/tensorrt/batched_nms/allClassNMS.cu
+++ b/csrc/backend_ops/tensorrt/batched_nms/allClassNMS.cu
@@ -5,6 +5,8 @@
 
 #include "kernel.h"
 
+#define CudaAssert( X ) if ( !(X) ) { printf( "Thread %d:%d failed assert at %s:%d!\n", blockIdx.x, threadIdx.x, __FILE__, __LINE__ ); return; }
+
 template <typename T_BBOX>
 __device__ T_BBOX bboxSize(const Bbox<T_BBOX> &bbox, const bool normalized, T_BBOX offset) {
   if (bbox.xmax < bbox.xmin || bbox.ymax < bbox.ymin) {
@@ -84,6 +86,10 @@ __global__ void allClassNMS_kernel(const int num, const int num_classes,
                                    bool flipXY = false) {
   //__shared__ bool kept_bboxinfo_flag[CAFFE_CUDA_NUM_THREADS * TSIZE];
   extern __shared__ bool kept_bboxinfo_flag[];
+
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ <=720
+  CudaAssert((top_k<201) && "pre_top_k in deploy_cfg should be reduced for devices with arch 7.2!\n" );
+#endif
 
   for (int i = 0; i < num; i++) {
     const int offset = i * num_classes * num_preds_per_class + blockIdx.x * num_preds_per_class;
--- a/csrc/backend_ops/tensorrt/batched_nms/allClassNMS.cu
+++ b/csrc/backend_ops/tensorrt/batched_nms/allClassNMS.cu
@@ -252,7 +252,8 @@ pluginStatus_t allClassNMS(cudaStream_t stream, const int num, const int num_cla
                            void *afterNMS_scores, void *afterNMS_index_array, bool flipXY) {
   auto __cuda_arch__ = get_cuda_arch(0);  // assume there is only one arch 7.2 device
   if (__cuda_arch__ == 720 && top_k >= 1000) {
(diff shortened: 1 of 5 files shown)


docs/en/tutorials/how_to_install_mmdeploy_on_jetsons.md
```diff
@@ -117,3 +117,11 @@ Then install onnx through:
 pip install onnx
 ```
 Then reinstall mmdeploy.
+
+
+### FAQs
+
+- For TX2, `#assertion/root/workspace/mmdeploy/csrc/backend_ops/tensorrt/batched_nms/trt_batched_nms.cpp,98 2022-02-08 07:15:09,334 - mmdeploy - ERROR - visualize tensorrt model failed.`
```
grimoire, line 124: Not just tx2, all devices with arch==7.2 have the same problem.

docs/en/tutorials/how_to_install_mmdeploy_on_jetsons.md
```diff
@@ -117,3 +117,11 @@ Then install onnx through:
 pip install onnx
 ```
 Then reinstall mmdeploy.
+
+
+### FAQs
+
+- For devices with arch==7.2 , `#assertion/root/workspace/mmdeploy/csrc/backend_ops/tensorrt/batched_nms/trt_batched_nms.cpp,98 2022-02-08 07:15:09,334 - mmdeploy - ERROR - visualize tensorrt model failed.`
```
grimoire, line 124: the user might not understand what is arch==7.2, use jetson tx2, and jetson nano instead.
And we can add an assert in the nms plugin if arch==7.2.

csrc/backend_ops/tensorrt/batched_nms/allClassNMS.cu
```diff
@@ -85,6 +87,10 @@ __global__ void allClassNMS_kernel(const int num, const int num_classes,
   //__shared__ bool kept_bboxinfo_flag[CAFFE_CUDA_NUM_THREADS * TSIZE];
   extern __shared__ bool kept_bboxinfo_flag[];
 
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ <=720
```
grimoire, line 90: can we move assert out of the kernel?

AllentDan: Seems we can't from [this blog](https://www.rookiehpc.com/cuda/docs/__cuda_arch__.php). While in this [blog](https://llvm.org/docs/CompileCudaWithLLVM.html), `__CUDA_ARCH__` will be defined if NVCC is using clang as a preprocessor. Maybe we can do the check for some threads like the first thread?

grimoire: @lvhan028 any idea?

lvhan028: I think we can try `static_assert`, which happens at compiling time

lvhan028: I am not sure whether it brings in `Undefined Behavior`

lzhangzz: The documentation for `__CUDA_ARCH__` says that host code should not depend on it. However, we can query the compute capability of the device using `cudaDeviceGetAttribute`

csrc/backend_ops/tensorrt/batched_nms/allClassNMS.cu
```diff
@@ -251,6 +250,11 @@ pluginStatus_t allClassNMS(cudaStream_t stream, const int num, const int num_cla
                            const bool isNormalized, const DataType DT_SCORE, const DataType DT_BBOX,
                            void *bbox_data, void *beforeNMS_scores, void *beforeNMS_index_array,
                            void *afterNMS_scores, void *afterNMS_index_array, bool flipXY) {
+  auto __cuda_arch__ = get_cuda_arch(0);  // assume there is only one arch 7.2 device
+  if (__cuda_arch__ == 720 && top_k > 200) {
```
grimoire, line 254: Why 200?

AllentDan: Just a number from experience.

grimoire: Too small for object detection task.

grimoire (approved):

grimoire:
Errr, I am not sure if adding an assert is a good idea if we do not known the max topk.
Maybe a warning is enough?

grimoire (approved):
LGTM

RunningLeon (approved):
LGTM

Issue #238, opened 2022-03-15
deploy.py crashes on headless installation


tehkillerbee:
Hello everyone,

I noticed a small bug when deploying on a headless installation.

The deployment process will terminate, before generating test output. It does appear that the correct engine output is generated, just not the validation image output.

A simple workaround is to run the deploy.py with xforwarding enabled.
```
2022-03-15:06:29:28,root ERROR    [utils.py:41] no display name and no $DISPLAY environment variable
Traceback (most recent call last):
  File "/home/nvidia/git/work/MMDeploy/mmdeploy/utils/utils.py", line 36, in target_wrapper
    result = target(*args, **kwargs)
  File "/home/nvidia/git/work/MMDeploy/mmdeploy/apis/visualize.py", line 68, in visualize_model
    show_result=show_result)
  File "/home/nvidia/git/work/MMDeploy/mmdeploy/codebase/mmdet/deploy/object_detection.py", line 184, in visualize
    score_thr=score_thr)
  File "/home/nvidia/git/work/MMDeploy/mmdeploy/codebase/mmdet/deploy/object_detection_model.py", line 286, in show_result
    out_file=out_file)
  File "/home/nvidia/git/work/mmdetection/mmdet/models/detectors/base.py", line 353, in show_result
    out_file=out_file)
  File "/home/nvidia/git/work/mmdetection/mmdet/core/visualization/image.py", line 119, in imshow_det_bboxes
    fig = plt.figure(win_name, frameon=False)
  File "/usr/lib/python3/dist-packages/matplotlib/pyplot.py", line 539, in figure
    **kwargs)
  File "/usr/lib/python3/dist-packages/matplotlib/backend_bases.py", line 171, in new_figure_manager
    return cls.new_figure_manager_given_figure(num, fig)
  File "/usr/lib/python3/dist-packages/matplotlib/backends/backend_tkagg.py", line 1049, in new_figure_manager_given_figure
    window = Tk.Tk(className="matplotlib")
  File "/usr/lib/python3.6/tkinter/__init__.py", line 2023, in __init__
    self.tk = _tkinter.create(screenName, baseName, className, interactive, wantobjects, useTk, sync, use)
_tkinter.TclError: no display name and no $DISPLAY environment variable
2022-03-15 06:29:30,000 - mmdeploy - ERROR - visualize tensorrt model failed.
```

hanrui1sensetime:
[...]

PR #250, opened 2022-03-18
[Fix] add support for headless device

hanrui1sensetime:

## Motivation

MMDeploy is failed on headless devices.(#238 )

## Modification

we check $DISPLAY for headless devices. If headless device, we don't show the visualization.

## BC-breaking (Optional)

None.

## Use cases (Optional)

On headless device, execute tools/deploy.py as [get_started](https://github.com/open-mmlab/mmdeploy/blob/master/docs/en/get_started.md)

## Checklist

### DIFF
--- a/tools/deploy.py
+++ b/tools/deploy.py
@@ -248,18 +248,19 @@ def main():
     if args.test_img is None:
         args.test_img = args.img
     import os
-    is_display = os.getenv("DISPLAY")
+    is_display = os.getenv('DISPLAY')
     # for headless installation.
     if is_display is not None:
         # visualize model of the backend
         create_process(
             f'visualize {backend.value} model',
             target=visualize_model,
-            args=(model_cfg_path, deploy_cfg_path, backend_files, args.test_img,
-                  args.device),
+            args=(model_cfg_path, deploy_cfg_path, backend_files,
+                  args.test_img, args.device),
             kwargs=dict(
                 backend=backend,
-                output_file=osp.join(args.work_dir, f'output_{backend.value}.jpg'),
+                output_file=osp.join(args.work_dir,
+                                     f'output_{backend.value}.jpg'),
                 show_result=args.show),
             ret_value=ret_value)
 
--- a/tools/deploy.py
+++ b/tools/deploy.py
@@ -247,29 +247,33 @@ def main():
 
     if args.test_img is None:
         args.test_img = args.img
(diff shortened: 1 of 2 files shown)


grimoire:
A better way is to disable visualize in deploy.py. Any advice @RunningLeon ?

RunningLeon:
> A better way is to disable visualize in deploy.py. Any advice @RunningLeon ?

Disable `visuallize` will add another argument to `deploy.py`, which makes it harder to use. Maybe it is better to handled inside mmdeploy.

hanrui1sensetime:
> > A better way is to disable visualize in deploy.py. Any advice @RunningLeon ?
> 
> Disable `visuallize` will add another argument to `deploy.py`, which makes it harder to use. Maybe it is better to handled inside mmdeploy.

I think we need not add another argument to `deploy.py` if we disable `visualize`, because for headless we only detect the $DISPLAY variable to work, which can call os package to finish.

grimoire (approved):
LGTM. Could add a warning when display is none.

tools/deploy.py
```diff
@@ -247,29 +247,36 @@ def main():
...
+            target=visualize_model,
+            args=(model_cfg_path, deploy_cfg_path, [checkpoint_path],
+                  args.test_img, args.device),
+            kwargs=dict(
+                backend=Backend.PYTORCH,
+                output_file=osp.join(args.work_dir, 'output_pytorch.jpg'),
+                show_result=args.show),
+            ret_value=ret_value)
+    else:
+        logger.warning('$DISPLAY is none.')
```
RunningLeon, line 279: warning message is too short to let user know what's it wrong here.

tools/deploy.py
```diff
@@ -247,29 +247,37 @@ def main():
...
+            target=visualize_model,
+            args=(model_cfg_path, deploy_cfg_path, [checkpoint_path],
+                  args.test_img, args.device),
+            kwargs=dict(
+                backend=Backend.PYTORCH,
+                output_file=osp.join(args.work_dir, 'output_pytorch.jpg'),
+                show_result=args.show),
+            ret_value=ret_value)
+    else:
+        logger.warning('$DISPLAY is none. \
```
RunningLeon, line 279: "visualize_model" has been skipped maybe because it's running on a headless device.

RunningLeon (approved):
LGTM when change the logging message

PR #251, opened 2022-03-18
[Enhancement]: Support fcn_unet deployment with dynamic shape

RunningLeon:
## Motivation
Support `fcn_unet` exporting with dynamic shape.  
Test config: `configs/unet/fcn_unet_s5-d16_4x4_512x1024_160k_cityscapes.py`

## Modification

- [x] Rewrite [UpConvBlock](https://github.com/open-mmlab/mmsegmentation/blob/1b24ad656f7c77bd79100bd7f35a00827043e53d/mmseg/models/utils/up_conv_block.py#L99) to support dynamically upsampling with `size` instead of `scale_factor`. 
- [x] Add unit test for rewriting of `UpConvBlock`
- [x] Update related docs

## BC-breaking (Optional)

None
## Use cases (Optional)

## Checklist


mmdeploy/codebase/mmseg/models/utils/up_conv_block.py
```diff
@@ -0,0 +1,37 @@
...
+        skip (Tensor): Skip branch feature.
+        x (Tensor): Input feature to be upsampled.
+
+    Returns:
+        Tensor: Upsampled output feature map.
+    """
+    if is_dynamic_shape(ctx.cfg):
+        # upsample with `size` instead of `scale_factor`
+        from mmseg.ops import Upsample
+        for c in self.upsample.interp_upsample:
```
grimoire, line 29: upsample cound be a `ConvModule`.https://github.com/open-mmlab/mmsegmentation/blob/98984bf8553a83bc4a2231127d4da9323ac4db27/mmseg/models/utils/up_conv_block.py#L85

Is it safe to loop over upsample.interp_upsample?

AllentDan (commented):
`mmdeploy/codebase/mmseg/models/utils/up_conv_block.py` may be a little deep? Shall we remove `utils`?

mmdeploy/codebase/mmseg/models/utils/up_conv_block.py
RunningLeon: maybe we could check type of self.upsample at first.

RunningLeon:
> `mmdeploy/codebase/mmseg/models/utils/up_conv_block.py` may be a little deep? Shall we remove `utils`?

I think it's OK since `utils` is in the same level as `decode_heads`

AllentDan (approved):

grimoire (approved):
LGTM

PR #271, opened 2022-03-24
[WIP Feature] Regression test for mmdeploy

PeterH0323:
## Motivation

Regression test for mmedploy.

## Modification

This PR is for mmdeploy regession test.
 
Here is the flow chart of my design:

<div align=center>

</div>

There is a example for deploy-yml :

image

This is a report example:

image

## Codebases & Metric
- [x] mmdet
  - [x] bbox
  - [ ] segm
  - [ ] PQ 
- [ ] mmcls
- [ ] mmdet3d
- [ ] mmedit
- [ ] mmocr
- [ ] mmpose
- [ ] mmseg

## Backends
- [x] onnxruntime
- [x] TensorRT
- [x] PPLNN
- [x] NCNN
- [x] OpenVino
- [ ] MMDeploy SDK

## Use cases (Optional)

```python
python ./tools/regression_test.py \
    --deploy-yml "${DEPLOY_YML_PATH}" \
    --test-type "${TEST_TYPE}" \
    --backend "${BACKEND}"  \
    --work-dir "${WORK_DIR}" \
    --device "${DEVICE}" \    
    --log-level INFO
```

## Checklist

- [ ] Pre-commit or other linting tools are used to fix the potential lint issues.
- [x] The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
- [ ] If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
- [x] The documentation has been modified accordingly, like docstring or example tutorials.


configs/mmdet/mmdet_regression_test.yaml
```diff
@@ -0,0 +1,175 @@
+globals:
+  codebase_name: mmdet
+  codebase_release: V2.22.0
```
RunningLeon, line 3: maybe we should remove version info and assume it is determined by user's env.

configs/mmdet/mmdet_regression_test.yaml
```diff
@@ -0,0 +1,175 @@
...
+          fp32: &default_pplnn_dynamic_fp32
+    sdk: &default_sdk_config
+      performance_align: False
+      deploy_config:
+        static:
+          fp32: &default_sdk_static_fp32
+          fp16: &default_sdk_static_fp16
+          int8: &default_sdk_static_int8
+        dynamic:
+          fp32: &default_sdk_dynamic_fp32
```
RunningLeon, line 61: we assume sdk only have one config, say dynamic: fp32: configs/mmdet/detection/detection_sdk_dynamic.py for detection

configs/mmdet/mmdet_regression_test.yaml
PeterH0323: Done

configs/mmdet/mmdet_regression_test.yaml
PeterH0323: Done

configs/mmdet/mmdet_regression_test.yaml
```diff
@@ -0,0 +1,170 @@
...
+      - yolov3_d53_320_273e_coco.py
+      - yolov3_mobilenetv2_320_300e_coco.py
+    metafile: configs/yolo/metafile.yml
+    deploy_config_dir: ./configs/mmdet/detection
+    calib_dataset_cfg: None
+    backends:
+      onnxruntime:
+        performance_align: False
+        deploy_config:
+          static:
```
RunningLeon, line 73: keep either static config or dynamic config. No need to keep both. For those support dynamic shape, keep dynamic config. For those only support static shape, keep static config.

RunningLeon: I'll try to update this config.

PeterH0323: OK, If some model do not need to test both type of shape, we can delete static or dynamic, the test script will only test backend config below `deploy_config`

tools/regression_test.py
```diff
@@ -0,0 +1,674 @@
...
+            backend_output_path = Path(work_dir). \
+                joinpath(Path(checkpoint_path).parent.parent.name,
+                         Path(checkpoint_path).parent.name,
+                         backend_name,
+                         infer_type,
+                         Path(checkpoint_path).stem)
+            backend_output_path = backend_output_path.absolute().resolve()
+            backend_output_path.mkdir(parents=True, exist_ok=True)
+
+            cmd_str = f'cd {str(Path().cwd())} && ' \
```
RunningLeon, line 445: here we should consider two things:
1. if it is a dynamic shape config, `--test-img` should be given with a different size image.
2. if it is for sdk, '--dumy-info` should be added.

tools/regression_test.py
```diff
@@ -0,0 +1,674 @@
...
+                      f'{str(checkpoint_path.absolute().resolve())} ' \
+                      './tests/data/tiger.jpeg ' \
+                      f'--work-dir {backend_output_path} ' \
+                      f'--device {device_type} ' \
+                      '--log-level INFO'
+
+            print(f'Process cmd = {cmd_str}')
+
+            # Convert the model to specific backend
+            os.system(cmd_str)
```
RunningLeon, line 458: what if running cmd failed, should we try-catch here?

tools/regression_test.py
PeterH0323: About `1`, I think I can add a field named `test_image_path` in test yaml, it is better?

tools/regression_test.py
PeterH0323: Done

tools/regression_test.py
RunningLeon: Ok

RunningLeon: ok

PeterH0323: > here we should consider two things:
> 
> 1. if it is a dynamic shape config, `--test-img` should be given with a different size image.
> 2. if it is for sdk, '--dumy-info` should be added.

All done

tools/regression_test.py
```diff
@@ -0,0 +1,686 @@
...
+
+    df = pd.DataFrame(report_info)
+    df.to_excel(report_save_path)
+
+    logger.info(f'Saved regression test report '
+                f'to {report_save_path}.')
+
+
+def main():
+    args = parse_args()
```
RunningLeon, line 577: does `args.test_type` work in this version? what's the result table look like if only test conversion?

tools/regression_test.py
```diff
@@ -0,0 +1,686 @@
...
+
+from mmcv.parallel import MMDataParallel
+from mmdeploy.apis import build_task_processor
+from mmdeploy.utils import load_config, get_root_logger, parse_device_id
+from mmdeploy.utils.timer import TimeCounter
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Process Regression Test')
+    parser.add_argument('--deploy-yml', help='regression test yaml path',
```
RunningLeon, line 21: add ` nargs='+'` if argumet receive multiple inputs

tools/regression_test.py
```diff
@@ -0,0 +1,686 @@
...
+                        default='./configs/mmdet/mmdet_regression_test.yaml')
+    parser.add_argument('--test-type', help='`test type', default='precision')
+    parser.add_argument('--backend', help='test specific backend(s)',
```
RunningLeon, line 24: is this argument used in this version?

tools/regression_test.py
PeterH0323: Done. It will show `-` in the `fps` and `metric`, but will show `True` when conver is success

tools/regression_test.py
PeterH0323: Done

tools/regression_test.py
PeterH0323: Done. In my least commit, I make them work.

tools/regression_test.py
```diff
@@ -0,0 +1,711 @@
...
+                         Path(checkpoint_path).stem)
+            backend_output_path = backend_output_path.absolute().resolve()
+            backend_output_path.mkdir(parents=True, exist_ok=True)
+
+            cmd_str = f'cd {str(Path().cwd())} && ' \
+                      'python3 ./tools/deploy.py ' \
+                      f'{str(deploy_cfg_path.absolute().resolve())} ' \
+                      f'{str(model_cfg_path.absolute().resolve())} ' \
+                      f'{str(checkpoint_path.absolute().resolve())} ' \
+                      './tests/data/tiger.jpeg ' \
```
RunningLeon, line 469: we may add input image and test image in global info

configs/mmdet/mmdet_regression_test.yaml
```diff
@@ -0,0 +1,182 @@
...
+    codebase_model_config_dir: ./configs/retinanet
+    model_configs:
+      - retinanet_r50_fpn_1x_coco.py
+    metafile: configs/retinanet/metafile.yml
+    deploy_config_dir: ./configs/mmdet/detection
+    input_img_path: *default_input_img_path
+    dynamic_test_img_path:
+    calib_dataset_cfg:
+    sdk_config:
+      deploy_config_dir: ./configs/mmdet/detection
```
RunningLeon, line 85: deploy_config_dir should be moved to global_info

PeterH0323: Done

PeterH0323:
I have some problem when I try to rebase the code. So I open another PR #302 and close this one.

Issue #268, opened 2022-03-24
about swin model deployment


Joll123:
I run ./tools/deploy.py expecting to export the .onnx model, I get an error. How to deal with it?

2022-03-23 17:48:30,339 - mmdeploy - INFO - torch2onnx start.
2022-03-23:17:48:42,root ERROR [utils.py:43] CascadeRCNN: SwinTransformer: init() got an unexpected keyword argument 'embed_dim'
Traceback (most recent call last):
File "/Users/jinhonglong/opt/anaconda3/envs/openmmlab/lib/python3.7/site-packages/mmcv/utils/registry.py", line 52, in build_from_cfg
return obj_cls(**args)
TypeError: init() got an unexpected keyword argument 'embed_dim'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
File "/Users/jinhonglong/opt/anaconda3/envs/openmmlab/lib/python3.7/site-packages/mmcv/utils/registry.py", line 52, in build_from_cfg
return obj_cls(**args)
File "/Users/jinhonglong/opt/anaconda3/envs/openmmlab/lib/python3.7/site-packages/mmdet/models/detectors/cascade_rcnn.py", line 28, in init
init_cfg=init_cfg)
File "/Users/jinhonglong/opt/anaconda3/envs/openmmlab/lib/python3.7/site-packages/mmdet/models/detectors/two_stage.py", line 32, in init
self.backbone = build_backbone(backbone)
File "/Users/jinhonglong/opt/anaconda3/envs/openmmlab/lib/python3.7/site-packages/mmdet/models/builder.py", line 20, in build_backbone
return BACKBONES.build(cfg)
File "/Users/jinhonglong/opt/anaconda3/envs/openmmlab/lib/python3.7/site-packages/mmcv/utils/registry.py", line 212, in build
return self.build_func(*args, **kwargs, registry=self)
File "/Users/jinhonglong/opt/anaconda3/envs/openmmlab/lib/python3.7/site-packages/mmcv/cnn/builder.py", line 27, in build_model_from_cfg
return build_from_cfg(cfg, registry, default_args)
File "/Users/jinhonglong/opt/anaconda3/envs/openmmlab/lib/python3.7/site-packages/mmcv/utils/registry.py", line 55, in build_from_cfg
raise type(e)(f'{obj_cls.name}: {e}')
TypeError: SwinTransformer: init() got an unexpected keyword argument 'embed_dim'

[...]

PR #272, opened 2022-03-24
[Bug] compatible with torch 1.11

AllentDan:
Use `verbose` instead of `strip_doc_string` for onnx export according to [official docstring](https://pytorch.org/docs/master/onnx.html#functions)

Related issue #268.

### DIFF
--- a/mmdeploy/apis/pytorch2onnx.py
+++ b/mmdeploy/apis/pytorch2onnx.py
@@ -32,6 +32,8 @@ def torch2onnx_impl(model: torch.nn.Module, input: torch.Tensor,
     output_names = onnx_cfg['output_names']
     axis_names = input_names + output_names
     dynamic_axes = get_dynamic_axes(deploy_cfg, axis_names)
+    verbose = not onnx_cfg.get('strip_doc_string', True) or onnx_cfg.get(
+        'verbose', False)
 
     # patch model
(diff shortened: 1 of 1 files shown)


grimoire:
I can't find any description about `strip_doc_string` in the official link you provided.

AllentDan:
> I can't find any description about `strip_doc_string` in the official link you provided.

Yes, but it is used in our codebase. It is deprecated in the latest pytorch.

AllentDan:
> verbose

Given by the userwarning of pytorch:
```
UserWarning: strip_doc_string' is deprecated and ignored. Will be removed in next PyTorch release. It's combined with verbose' argument now. warnings.warn("strip_doc_string' is deprecated and ignored. Will be removed in "
```

grimoire:
Do we really need this doc_string? Can we simply remove it? Verbose is really annoy.

AllentDan:
Well, that makes no difference to me acutally. But we may ask @SemyonBevzuk for the reason adding it and detemine then.

SemyonBevzuk:
Previously, `strip_doc_string` could be useful for determining which string in Python code is associated with a graph node in ONNX.
This can be quite useful when debugging export-related errors. So it seems to me that this functionality needs to be supported.

grimoire (approved):

SemyonBevzuk (approved):

PR #273, opened 2022-03-24
Reduce torch dependency

grimoire:
This PR aims to reduce torch dependency when exporting TensorRT.
Note that we need `pycuda` as a extra dependency to handle device change and cudaMemcpy. I am not sure if we should add it into requirements.

This PR is based on master, we can merge it after v0.4.0.

Note that we still can not use it without PyTorch since `__init__.py` would import those unrelated module with torch imported.


grimoire:
Close this, it would be update in https://github.com/open-mmlab/mmdeploy/pull/347

PR #286, opened 2022-03-29
[Doc] update benchmark add supported-model-list

AllentDan:
Update as it was found that cascade MRCNN can be exported to TensorRT. And after we set `export_postprocess_mask=False` as default, MRCNN benchmark on T4 should be updated as well.


grimoire (approved):
LGTM

PR #310, opened 2022-04-04
fixed a bug causes text-recognizer to fail when (non-NULL) empty bbox list is passed

lzhangzz:
Inside `mmdeploy_text_recognizer_apply_bbox`, when `bbox_count` is 0 for an image, inference will be launched with no input supplied.

This PR resolves the problem by skipping inference for images with `bbox_count` equal to zero.

### DIFF
--- a/csrc/apis/c/text_recognizer.cpp
+++ b/csrc/apis/c/text_recognizer.cpp
@@ -108,15 +108,21 @@ int mmdeploy_text_recognizer_apply_bbox(mm_handle_t handle, const mm_mat_t *imag
 
   try {
     auto recognizer = static_cast<Handle *>(handle);
-    Value input{Value::kArray, Value::kArray};
+    Value::Array input_images;
+    Value::Array input_bboxes;
     auto _bboxes = bboxes;
     auto result_count = 0;
-    for (int i = 0; i < image_count; ++i) {
-      mmdeploy::Mat _mat{images[i].height,       images[i].width, PixelFormat(images[i].format),
-                         DataType(images->type), images[i].data,  Device{"cpu"}};
-      input[0].push_back({{"ori_img", _mat}});
 
+    // mapping from image index to result index, -1 represents invalid image with no bboxes
+    // supplied.
+    std::vector<int> result_index(image_count, -1);
+
+    for (int i = 0; i < image_count; ++i) {
       if (bboxes && bbox_count) {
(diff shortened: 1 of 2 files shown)


csrc/apis/c/text_recognizer.cpp
```diff
@@ -108,15 +108,21 @@ int mmdeploy_text_recognizer_apply_bbox(mm_handle_t handle, const mm_mat_t *imag
...
 
+    // mapping from image index to result index, -1 represents invalid image with no bboxes
+    // supplied.
+    std::vector<int> result_index(image_count, -1);
+
+    for (int i = 0; i < image_count; ++i) {
       if (bboxes && bbox_count) {
+        if (bbox_count[i] == 0) {
+          // skip images with no bounding boxes (push nothing)
+          continue;
```
lvhan028, line 124: It looks `continue` might result in the mismatch beween `input_images` and `input_bboxes`, doesn't it?

csrc/apis/c/text_recognizer.cpp
```diff
@@ -128,17 +134,26 @@ int mmdeploy_text_recognizer_apply_bbox(mm_handle_t handle, const mm_mat_t *imag
...
-        input[1].push_back({{"boxes", boxes}});
+        input_bboxes.push_back({{"boxes", boxes}});
       } else {
-        input[1].push_back(Value::kNull);
+        // bboxes or bbox_count not supplied, use whole image
         result_count += 1;
+        input_bboxes.push_back(Value::kNull);
       }
+
+      result_index[i] = static_cast<int>(input_images.size());
```
lvhan028, line 144: All the entries have the same value?

lzhangzz: No, it's not.

lzhangzz: No, they are not.

csrc/apis/c/text_recognizer.cpp
lvhan028: My mistake

csrc/apis/c/text_recognizer.cpp
lvhan028: My mistake

lvhan028 (approved):

irexyc (approved):

Issue #260, opened 2022-03-22
Regarding supported version of spdlog


tehkillerbee:
Hello, 

I've had some issues building and running MMDeploy with older versions of spdlog (`libspdlog-dev/bionic 1:0.16.3-1 arm64
`, as installed by apt-get on Jetson). I had to update to the latest version on spdlog from git to solve the issue. It appears that MMDEPLOY_INFO, _ERROR messages result in a crash/segfault whenever I build with the old spdlog.

Are there any known compatibility issues with older versions of spdlog? I do not recall having this issue before upgrading to MMdeploy 0.3.0. 

Thanks!

lzhangzz:
Hi,  

Sorry for the late reply. 

There is some adjustments on the project structure in v0.3.0, which may introduce some side-effects on linkage of some components.

Can provide more info about your settings and the excat cmake command used to configure the project so we can reproduce the problem?

tehkillerbee:
Sure!

I previously installed libspdlog-dev using the package manager, as instructed in the user guide. On latest Ubuntu 20.04, this installs 1:1.5.0-1 while on 18.04 (eg. Jetson), this installs 1:0.16.3-1.

I then built ppl.cv and MMDeploy as usual

```bash
SPDLOG_DIR=${WORKING_DIR}/spdlog
PPLCV_DIR=${WORKING_DIR}/ppl.cv
MMDEPLOY_DIR=${WORKING_DIR}/MMDeploy

## install from package manager
sudo apt-get install libspdlog-dev

## pplcv
cd ${WORKING_DIR}
git clone https://github.com/openppl-public/ppl.cv.git ${PPLCV_DIR}
cd ${PPLCV_DIR}
git checkout tags/v0.6.2
./build.sh cuda

## mmdeploy SDK
cd ${WORKING_DIR}
git clone https://github.com/open-mmlab/mmdeploy.git ${MMDEPLOY_DIR}
cd ${MMDEPLOY_DIR}
git checkout tags/v0.3.0
git submodule update --init --recursive
pip install -e .

mkdir build -p && cd build
cmake .. \
  -DMMDEPLOY_BUILD_SDK=ON \
  -DCMAKE_CXX_COMPILER=g++ \
  -Dpplcv_DIR=${PPLCV_DIR}/cuda-build/install/lib/cmake/ppl \
  -DMMDEPLOY_TARGET_DEVICES="cuda;cpu" \
[...]

PR #313, opened 2022-04-06
[Fix] include missing <type_traits> for formatter.h

lzhangzz:
#260

### DIFF
--- a/csrc/core/utils/formatter.h
+++ b/csrc/core/utils/formatter.h
@@ -38,7 +38,7 @@ inline void format_arg(BasicFormatter<char> &f, const char *, const mmdeploy::Va
   f.writer() << mmdeploy::format_value(d);
 }
 
-template <typename T, std::enable_if_t<std::is_enum_v<std::decay_t<T> >, bool> = true>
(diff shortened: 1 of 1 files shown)


irexyc (approved):

lvhan028:
@tehkillerbee Could you please review this PR?

tehkillerbee:
@lvhan028 Yep I will do it right away, I needed to get access to the Jetson AGX to test with the old spdlog/fmt version

tehkillerbee:
@lvhan028 I get the same error as before; 
```
/usr/local/include/cpp/core/utils/formatter.h:40:45: error: ‘is_enum_v’ is not a member of ‘std’
 template <typename T, std::enable_if_t<std::is_enum_v<std::decay_t<T> >, bool> = true>
```

It looks like the main issue is  `is_enum_v` template is only available with C++17 and above. It is possible we do not need to include #include <type_traits> after all.  If you build the examples with `-DCMAKE_CXX_STANDARD=17` it builds fine.

A workaround could be to use `is_enum` directly. That seems to work correctly.
```
template <typename T, std::enable_if_t<std::is_enum<std::decay_t<T> >::value, bool> = true>
```

tehkillerbee (approved):

PR #331, opened 2022-04-11
[Fix]fix pillarencode

VVsssssk:
## Motivation

Fix pillar_encode rewrite func. The original function has redundant calculation, and I modified the function to make the function calculation smaller.

## BC-breaking (Optional)

None.


RunningLeon:
@VVsssssk pls. update PR description and clarify why should do this change.

grimoire (approved):
LGTM

RunningLeon (approved):
LGTM

PR #336, opened 2022-04-12
[Fix] MMEditing cannot save results when testing

SingleZombie:
## Motivation

We can save results when using `tools/test.py`. However, MMEditing fails to do this because we lack some codes.

## Modification

Copy the codes related to save results from `mmedit.basic_restorer.py`  to `mmdeploy.super_resolution_model.py`

### DIFF
--- a/mmdeploy/codebase/mmedit/deploy/super_resolution_model.py
+++ b/mmdeploy/codebase/mmedit/deploy/super_resolution_model.py
@@ -90,7 +90,6 @@ class End2EndModel(BaseBackendModel):
                      lq: torch.Tensor,
                      gt: Optional[torch.Tensor] = None,
                      meta=None,
-                     save_image=False,
                      save_path=None,
                      *args,
                      **kwargs):
@@ -100,7 +99,6 @@ class End2EndModel(BaseBackendModel):
             lq (torch.Tensor): The input low-quality image of the model.
             gt (torch.Tensor): The ground truth of input image. Defaults to
                 `None`.
-            save_image (bool): Whether to save image. Default: False.
             save_path (str): Path to save image. Default: None.
             *args: Other arguments.
             **kwargs: Other key-pair arguments.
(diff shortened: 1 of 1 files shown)


mmdeploy/codebase/mmedit/deploy/super_resolution_model.py
```diff
@@ -88,6 +89,9 @@ def forward(self,
     def forward_test(self,
                      lq: torch.Tensor,
                      gt: Optional[torch.Tensor] = None,
+                     meta=None,
```
grimoire, line 92: Please add type hint.

mmdeploy/codebase/mmedit/deploy/super_resolution_model.py
```diff
@@ -96,6 +100,8 @@ def forward_test(self,
             lq (torch.Tensor): The input low-quality image of the model.
             gt (torch.Tensor): The ground truth of input image. Defaults to
                 `None`.
+            save_image (bool): Whether to save image. Default: False.
```
grimoire, line 103: Where is the root docstring of `meta`?

mmdeploy/codebase/mmedit/deploy/super_resolution_model.py
```diff
@@ -104,6 +110,17 @@ def forward_test(self,
         """
         outputs = self.forward_dummy(lq)
         result = self.test_post_process(outputs, lq, gt)
+
+        # Align to mmediting BasicRestorer
+        if save_image and save_path:
```
AllentDan, line 115: seems `save_image` and `save_path` both need to be specified if I want to save. Keeping only `save_path` is enough then.

SingleZombie: I agree. Our `tools/test.py` uses `--show` and `--show-dir` to control the output behavior. Maybe it should also be simplified later.

AllentDan (approved):
LGTM after adding type hint.

grimoire (approved):

PR #347, opened 2022-04-14
[WIP]fix for pre-build

grimoire:

## Motivation

Please describe the motivation of this PR and the goal you want to achieve through this PR.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


grimoire:
Remove dependency in `mmdeploy/codebase` is not necessary. Since all these codebase reqire MMCV.

grimoire:
Note that new api might require some features from this PR.

mmdeploy/backend/tensorrt/utils.py
```diff
@@ -54,8 +52,17 @@ def create_trt_engine(onnx_model: Union[str, onnx.ModelProto],
         >>>             device_id=0)
         >>>             })
     """
+
+    import os
+    old_cuda_device = os.environ.get('CUDA_DEVICE', None)
+    os.environ['CUDA_DEVICE'] = str(device_id)
+    import pycuda.autoinit  # noqa:F401
```
RunningLeon, line 59: we may add pycuda to requirements

PeterH0323: And I tested on Jetson, I only install successful with `pycuda==2019.1`. I haven't look up why yet. 
I suggest we do not using lastest version of pycuda, we can set it to `2019.1`

RunningLeon: Found that pycuda has been released not so often. Have you tried version of 2021.1?

grimoire:
The builder has been moved here.

mmdeploy/utils/logging.py
```diff
@@ -0,0 +1,61 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import logging
+
+logger_initialized = {}
+
+
+def get_logger(name, log_file=None, log_level=logging.INFO, file_mode='w'):
```
RunningLeon, line 7: may add type hint for arguments

tools/package_tools/configs/linux_x86_trt.yaml
```diff
@@ -0,0 +1,24 @@
...
+    MMDEPLOY_BUILD_SDK: "ON"
+    MMDEPLOY_BUILD_SDK_PYTHON_API: "ON"
+    MMDEPLOY_TARGET_DEVICES: '"cpu;cuda"'
+    MMDEPLOY_TARGET_BACKENDS: "trt"
+    pplcv_DIR: ${pplcv_DIR}/cuda-build/install/lib/cmake/ppl
+
+local_configs:
+  - BUILD_NAME: "linux-x86-cuda102-trt7216"
+    cmake_envs:
+      TENSORRT_DIR: "${HOME}/SDK/TensorRT-7.2.1.6"
```
RunningLeon, line 13: could remove `/SDK` and users do not have to create the directory.

tools/package_tools/mmdeploy_builder.py
```diff
@@ -0,0 +1,244 @@
...
+        cfgs = yaml.load(f, yaml.Loader)
+
+    global_cfg = cfgs.get('global_config', dict())
+    local_cfgs = cfgs.get('local_configs', [])
+
+    merged_cfgs = [
+        _merge_cfg(global_cfg, local_cfg) for local_cfg in local_cfgs
+    ]
+
+    # if len(local_cfgs) == 0:
```
RunningLeon, line 220: could we remove these commented lines?

tools/package_tools/configs/windows_ort.yaml
```diff
@@ -0,0 +1,13 @@
...
+  cmake_flags: ['-G "Visual Studio 17 2022" -A x64 -T v143']
+  cmake_envs:
+    MMDEPLOY_BUILD_SDK: "ON"
+    MMDEPLOY_BUILD_SDK_PYTHON_API: "ON"
+    MMDEPLOY_TARGET_DEVICES: '"cpu"'
+    MMDEPLOY_TARGET_BACKENDS: "ort"
+    MMDEPLOY_CODEBASES: "all"
+
+local_configs:
+  - BUILD_NAME: "windows-cpu-ort8230"
```
lvhan028, line 11: Does `ort8230` mean onnxruntime 8.2.3.0? If so, the version is not correct

tools/package_tools/configs/windows_ort.yaml
```diff
@@ -0,0 +1,13 @@
+global_config:
```
lvhan028, line 1: Can we also provide windows-trt.yaml?

lvhan028:
Could you provide instructions about how to pack mmdeploy?

grimoire:
@lvhan028 All yamls in this are template configs. We need to decide which combination we need before adding the final config.

tools/package_tools/mmdeploy_builder.py
```diff
@@ -0,0 +1,236 @@
...
+    if sys.platform == 'win32':
+        # build cmd
+        build_cmd = 'cmake --build . --config Release -- /m'
+        _call_command(build_cmd, build_dir)
+        install_cmd = 'cmake --install . --config Release'
+        _call_command(install_cmd, build_dir)
+    else:
+        # build cmd
+        build_cmd = 'cmake --build . -- -j$(nproc) && cmake --install .'
+        _call_command(build_cmd, build_dir)
```
RunningLeon, line 119: when build fails here, should raise error and no need to build wheel

tools/package_tools/mmdeploy_builder.py
```diff
@@ -56,7 +56,14 @@ def _call_command(cmd, cwd, stdout=None, stderr=None):
         return
     logging.info(f'Process cmd: {cmd}')
     logging.info(f'work_path: {cwd}')
-    run(cmd, stdout=stdout, stderr=stderr, cwd=cwd, shell=True)
+    try:
+        ret = run(cmd, stdout=stdout, stderr=stderr, cwd=cwd, shell=True)
+        if ret.returncode != 0:
+            logging.error(f'Process cmd: "{cmd}"'
+                          f' failed with returncode: {ret.returncode}')
```
RunningLeon, line 63: should exit when `ret.returncode != 0`

tools/package_tools/mmdeploy_builder.py
```diff
@@ -56,7 +56,14 @@ def _call_command(cmd, cwd, stdout=None, stderr=None):
...
+    except Exception:
```
RunningLeon, line 64: error from subprocess cannot be catched here.

mmdeploy/backend/sdk/__init__.py
```diff
@@ -18,12 +18,15 @@
 
 if lib_path:
     lib_dir = os.path.dirname(lib_path)
-    sys.path.insert(0, lib_dir)
-
-if importlib.util.find_spec(module_name) is not None:
-    from .wrapper import SDKWrapper
-    __all__ = ['SDKWrapper']
-    _is_available = True
+    sys.path.append(lib_dir)
```
RunningLeon, line 21: should we update paths line 13-14 so that lib can be found?

grimoire: If you want to use sdk for pre-built package, you should install `mmdeploy_python` from the sdk package.

tools/package_tools/mmdeploy_builder.py
grimoire: Is it possible to catch the exception throw from the subprocess of commandline?

mmdeploy/backend/sdk/__init__.py
grimoire: It is not cause by `mmdeploy_python`,`SDKWrapper` require PyTorch...

tools/package_tools/mmdeploy_builder.py
RunningLeon: I've tested the try-catch won't work when building fails.

mmdeploy/backend/sdk/__init__.py
RunningLeon: I have PyTorch in my env and pytorch2onnx, onnx2tensorrt works fine, but running `test.py` with for sdk failed.

tools/package_tools/mmdeploy_builder.py
grimoire: Errr, I means it is impossible to catch an exception from `subprocess.run`.

mmdeploy/backend/sdk/__init__.py
grimoire: Try the latest commit.
Will `python -m 'import mmdeploy_python'` work?

tools/package_tools/mmdeploy_builder.py
RunningLeon: OK

mmdeploy/backend/sdk/__init__.py
RunningLeon: now it worked.

lvhan028:
```
CMake Error at cmake/modules/FindCUDNN.cmake:18 (message):
  Couldn't find cuDNN in CUDNN_DIR: , or in CUDA_TOOLKIT_ROOT_DIR:
  /usr/local/cuda-11.1, please check if the path is correct.
Call Stack (most recent call first):
  cmake/tensorrt.cmake:4 (include)
  csrc/backend_ops/tensorrt/CMakeLists.txt:7 (include)
```
`CUDNN_DIR` is not specified in the config file

lvhan028:
The name of the wheel file does not obey [PEP 425](https://peps.python.org/pep-0425/).
According to PEP 425, `mmdeploy-0.5.0-py2.py3-none-any.whl` misses platform tag. Besides, there is no python tag like `py2.py3`.
And mmdeploy SDK python package should be named as `mmdeploy_python-0.5.0-cp37-none-linux_x86_64.whl`

AllentDan (commented):
Tested fine with trt in my linux.

lvhan028:
```shell
.
├── dist
│   └── mmdeploy-0.5.0-py2.py3-none-any.whl
├── sdk-tar
│   ├── example
│   ├── include
│   ├── lib
│   └── python
└── sdk-tar.tar
```
I think there is no need to put `sdk-tar.tar` in the package. And it's better to rename `sdk-tar` as `sdk`.
After the package is built, we can compress it to `tar.gz` for Linux and `zip` for Windows, perspectively.
Have you had any idea about the package naming rules?

grimoire:
>Have you had any idea about the package naming rules?

Nope.
Should we use different package names for the different backend?

RunningLeon:
> Tested fine with trt in my linux.

Me too.

lvhan028:
> > Have you had any idea about the package naming rules?
> 
> Nope. Should we use different package names for the different backend?
I am afraid the package will explode if we put backend as a part of the package name.
Here is my proposal: 
mmdeploy-{os tag}-{cpu arch}-{device}?-version.{file extension}
e.g., 
- mmdeploy-linux-x86_64-gpu-0.5.0.tar.gz
- mmdeploy-windows-x86_64-0.5.0.zip

lvhan028:
| Mode   | LastWriteTime | Length Name |                  |
|--------|---------------|-------------|------------------|
| d----- | 5/17/2022     | 11:48 AM    | .mmdeploy_python |
| d----- | 5/17/2022     | 11:48 AM    | dist             |
| d----- | 5/17/2022     | 11:48 AM    | sdk              |

In windows, after running the prebuild scripts, a `.mmdeploy_python` is generated. Is it necessary? What is it used for?

grimoire:
> Mode	LastWriteTime	Length Name	
> d-----	5/17/2022	11:48 AM	.mmdeploy_python
> d-----	5/17/2022	11:48 AM	dist
> d-----	5/17/2022	11:48 AM	sdk
> In windows, after running the prebuild scripts, a `.mmdeploy_python` is generated. Is it necessary? What is it used for?

It is a copy of `tools/package_tools/packaging`. We need these codes to build `mmdeploy_python` wheel.

tools/package_tools/configs/windows_ort.yaml
```diff
@@ -0,0 +1,15 @@
...
+  bdist_tags:
+    python_tag: "py37"
```
irexyc, line 3: The python version in my env is python3.8, but the generated whl package name is mmdeploy_python-0.5.0-py37-none-win_amd64.whl.

tools/package_tools/configs/windows_ort.yaml
```diff
@@ -0,0 +1,15 @@
...
+  cmake_flags: ['-G "Visual Studio 17 2022" -A x64 -T v143']
+  cmake_envs:
```
irexyc, line 5: On my machine, I should add OpenCV_DIR and spdlog_DIR when building the windows_ort.yaml, or it can't find these library.

lvhan028:
> > Mode	LastWriteTime	Length Name
> > d-----	5/17/2022	11:48 AM	.mmdeploy_python
> > d-----	5/17/2022	11:48 AM	dist
> > d-----	5/17/2022	11:48 AM	sdk
> > In windows, after running the prebuild scripts, a `.mmdeploy_python` is generated. Is it necessary? What is it used for?
> 
> It is a copy of `tools/package_tools/packaging`. We need these codes to build `mmdeploy_python` wheel. Should I remove it after packaging?

If they are not necessary for our users, I think we can remove it safely.

tools/package_tools/configs/windows_ort.yaml
grimoire: Sure. I will try to read it from the environment.

lvhan028 (dismissed):

tools/package_tools/configs/jetson.yaml
```diff
@@ -10,4 +10,4 @@ global_config:
     pplcv_DIR: ${pplcv_DIR}/cuda-build/install/lib/cmake/ppl
 
 local_configs:
-  - BUILD_NAME: "mmdeploy-{version}-jetson"
```
lvhan028: maybe {version} -> {mmdeploy_v} ?

tools/package_tools/mmdeploy_builder.py
```diff
@@ -0,0 +1,354 @@
...
+    nvcc_cmd = 'nvcc' if len(CUDA_TOOLKIT_ROOT_DIR) <= 0 else osp.join(
+        CUDA_TOOLKIT_ROOT_DIR, 'bin', 'nvcc')
+
+    if osp.exists(nvcc_cmd):
+        nvcc = check_output(f'"{nvcc_cmd}" -V', shell=True)
+        nvcc = nvcc.decode('utf-8').strip()
+        pattern = r'Cuda compilation tools, release (\d+.\d+)'
+        match = re.search(pattern, nvcc)
+        if match is not None:
+            cuda_version = match.group(1)
```
irexyc, line 201: On windows, there is no CUDA_TOOLKIT_ROOT_DIR variable in windows_x64.xml, so it will use nvcc_cmd = 'nvcc'. It doesn't matter because the installation of cuda will add the nvcc bin folder to the path. 

While check osp.exists(nvcc_cmd) will fail on windows for two reasons: one is the nvcc_cmd is not a absoluate path, the other is that the executable file should be nvcc.exe other than nvcc.  May use try except there.

irexyc (approved):

Issue #340, opened 2022-04-13
windows + openvino can't ReadNetwork retinanet 


irexyc:
**Environment**

Windows10
onnxruntime 1.8.0
openvino openvino_2022.1.0.643

```
pip list
...
openvino                2022.1.0
openvino-dev            2022.1.0
...
```

**Reproduction**

1. convert pytorch to openvino
```
python tools/deploy.py ./configs/mmdet/detection/detection_openvino_dynamic-800x1344.py \ 
../mmdetection/configs/retinanet/retinanet_r50_fpn_1x_coco.py \ 
benchmark/retinanet_r50_fpn_1x_coco_20200130-c2398f9e.pth ../mmdetection/demo/demo.jpg \
--device cuda --dump-info --work-dir benchmark/openvino_detection_tmp_model 
```
2. create network
```
InferenceEngine::Core core_ = InferenceEngine::Core();
InferenceEngine::CNNNetwork network_ = core_.ReadNetwork(tmp_xml, tmp_bin);
```
error is :
```
[2022-04-14 15:48:32.304] [mmdeploy] [error] [openvino_net.cpp:136] unhandled exception when creating OpenVINO: Check 'PartialShape::broadcast_merge_into(pshape, node->get_input_partial_shape(i), autob)' failed at C:\j\workspace\private-ci\ie\build-windows-vs2019@3\b\repos\openvino\src\core\src\op\util\elementwise_args.cpp:30:
While validating node 'v1::Divide Divide_610 (Convert_3027[0]:f32{2}, Convert_3026[0]:f32{4}) -> ()' with friendly_name 'Divide_610':
Argument shapes are inconsistent.

[2022-04-14 15:48:32.313] [mmdeploy] [error] [openvino_net.cpp:265] error creating OpenVINONet: unknown (6) @ :0
[2022-04-14 15:48:32.314] [mmdeploy] [error] [detector.cpp:58] exception caught: unknown (6) @ :0
failed to create detector, code: 6
```

3. you can download my onnx and converted openvino model from here:
```
https://1drv.ms/u/s!Aqis6w3rjKXSh150FQr5Yu76TpqL?e=eSASXB
```

irexyc:
It is a code bug. #353 

The saved internal openvino model not in binary mode

PR #353, opened 2022-04-14
fix openvino export tmp model, add binary flag

irexyc:
## Motivation

1. the internal openvino not exported in binary and can't be loaded on windows (#340) 

## Modification

- [x] csrc/net/openvino/openvino_net.cpp

### DIFF
--- a/csrc/net/openvino/openvino_net.cpp
+++ b/csrc/net/openvino/openvino_net.cpp
@@ -86,10 +86,10 @@ Result<void> OpenVINONet::Init(const Value& args) {
   OUTCOME_TRY(auto raw_bin, model.ReadFile(config.weights));
 
   try {
-    std::ofstream xml_out(tmp_xml);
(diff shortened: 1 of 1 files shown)


cmake/modules/FindTENSORRT.cmake
```diff
@@ -9,22 +9,24 @@ endif()
 
 find_path(
     TENSORRT_INCLUDE_DIR NvInfer.h
-    HINTS ${TENSORRT_DIR}
```
lzhangzz, line 12: why there are TRT related stuff in a OpenVINO PR？

lzhangzz (changes requested):

cmake/modules/FindTENSORRT.cmake
irexyc: removed

lzhangzz (approved):
LGTM

lvhan028 (approved):

Issue #351, opened 2022-04-14
image_segmentation.cpp call onnxruntime-cuda thrown an abnormal


joryoker:
win10 mmdeploy image_segmentation.cpp call onnxruntime-static-cuda and onnxruntime-dynamic-cuda thrown an abnormal
call onnxruntime-static-cpu and onnxruntime-dynamic-cpu is ok
call tensorrt-static-cuda is ok
call tensorrt-dynamic-cuda error

![1649925571(1)](https://user-images.githubusercontent.com/45837432/163348480-755afba3-de9c-4e76-b989-371dd33c8619.png)
![1649925671(1)](https://user-images.githubusercontent.com/45837432/163348517-d6637dd9-6c56-4fd4-8e6c-651a3ffe7670.png)
![1649925777(1)](https://user-images.githubusercontent.com/45837432/163348536-39e2f89d-13eb-4f88-93f7-addf1027d451.png)
![1649925905(1)](https://user-images.githubusercontent.com/45837432/163348701-810c3aa1-e54e-4f55-b7d7-8719bf864027.png)

irexyc:
For tensorrt-dynamic-cuda, according to segmentation_tensorrt_dynamic-512x1024-2048x2048.py, the minimum shape of width is 1024, and you supply 512, so as the log said, your input is not corrent.
```
input_shapes=dict(
    input=dict(
        min_shape=[1, 3, 512, 1024],
        opt_shape=[1, 3, 1024, 2048],
        max_shape=[1, 3, 2048, 2048])))
```

For onnxruntime cuda, what onnxruntime version are you using? I will check it on my machine.

joryoker:
> For tensorrt-dynamic-cuda, according to segmentation_tensorrt_dynamic-512x1024-2048x2048.py, the minimum shape of width is 1024, and you supply 512, so as the log said, your input is not corrent.

thanks for you answer!
my onnxruntime version is 1.8.1

irexyc:
Found the reason. There is a bug in mmseg postprocess, should be fixed with PR363

joryoker:
i switch the csrc/codebase/mmseg，compile the mmdeploy SDK, but the same abnormal throwned.

> Found the reason. There is a bug in mmseg postprocess, should be fixed with PR363

irexyc:
[...]

PR #361, opened 2022-04-15
fix wrong buffer which will case openvino-gpu crash with segmentaion

irexyc:
## Motivation

address #351 

## Modification

- [x] csrc/codebase/mmseg/segment.cpp


lzhangzz:
It's better to modify the title to reflect the fact that this is a fix for all backends returning 64-bit integer class index, not limited to openvino, since #351 didn't mention anything about openvino.

 .

PR #362, opened 2022-04-15
fix wrong input mat type

irexyc:
## Motivation

In sdk c api, the input mat type of some codebase always point to the first mat type, which I think is wrong.

## Modification

- [x] csrc/apis/c/classifier.cpp
- [x] csrc/apis/c/detector.cpp
- [x] csrc/apis/c/pose_detector.cpp
- [x] csrc/apis/c/segmentor.cpp
- [x] csrc/apis/c/text_recognizer.cpp


lzhangzz (approved):
LGTM

lvhan028 (approved):

PR #363, opened 2022-04-15
fix wrong buffer which will case onnxruntime-gpu crash with segmentaion

irexyc:
## Motivation

address #351 

## Modification

- [x] csrc/codebase/mmseg/segment.cpp

### DIFF
--- a/csrc/codebase/mmseg/segment.cpp
+++ b/csrc/codebase/mmseg/segment.cpp
@@ -42,7 +42,7 @@ class ResizeMask : public MMSegmentation {
       // change kINT64 to 2 INT32
       TensorDesc desc{
           host_tensor.device(), DataType::kINT32, {1, 2, height, width}, host_tensor.name()};
-      Tensor _host_tensor(desc, mask.buffer());
+      Tensor _host_tensor(desc, host_tensor.buffer());
       return MaskResize(_host_tensor, input_height, input_width);
     } else if (mask.data_type() == DataType::kINT32) {
       return MaskResize(host_tensor, input_height, input_width);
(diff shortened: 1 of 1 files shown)


csrc/codebase/mmseg/segment.cpp
```diff
@@ -68,11 +68,20 @@ class ResizeMask : public MMSegmentation {
...
       cv::Mat _dst;
-      cv::extractChannel(dst, _dst, 0);
+      int channel = IsLittleEndian() ? 0 : dst.dims - 1;
+      cv::extractChannel(dst, _dst, channel);
       auto output_tensor = cpu::CVMat2Tensor(_dst);
       SegmentorOutput output{output_tensor, dst_height, dst_width, classes_};
       return to_value(output);
     }
+
+    bool IsLittleEndian() {
```
lzhangzz, line 78: Where is this comming from?

irexyc: updated

lzhangzz: let's convert it to a const boolean member since endianness will stay const for a single setting.

lzhangzz (approved):
LGTM

lvhan028 (approved):

lzhangzz (changes requested):
actually this can be done at compile time
```C++
constexpr bool is_little_endian() {
  constexpr union Un {
    char a;
    int b;
  } test {1};
  return (int) test.a == 1;
}
```
then use the constexpr function directly, or store it as a static constexpr member.

irexyc:
> actually this can be done at compile time
> 
> ```c++
> constexpr bool is_little_endian() {
>   constexpr union Un {
>     char a;
>     int b;
>   } test {1};
>   return (int) test.a == 1;
> }
> ```
> 
> then use the constexpr function directly, or store it as a static constexpr member.

Should change the order of variable of a and b.
But the compiler says it is not a constant expression [read of member 'a' of union with active member 'b' is not allowed in a constant expression](https://stackoverflow.com/questions/4239993/determining-endianness-at-compile-time#comment58835376_24431286)

lzhangzz (approved):

PR #368, opened 2022-04-15
update the circleci config file by adding workflows both for linux, windows and linux-gpu

lvhan028:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily receiving feedbacks. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

Add more tests in circleci

## Modification

- [x] test mmdeploy on ubuntu cpu
- [x] test mmdeploy on ubuntu cuda114
- [x] test mmdeploy on windows cpu

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
3. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #399, opened 2022-04-24
MMdeploy build & install script + prerequisites

tehkillerbee:
Hello, 

Over the last couple of months, I have been working on a bash script for building and installing MMDeploy and prerequisites to ease the install procedure when we use MMDetection, MMDeploy tools. Maybe this script can serve as a starting point for an easier install procedure when building and installing MMDeploy from source.

Naturally, this script is a work in progress. There are probably parts that can be cleaned up - and maybe some dependencies are not needed to install anymore.

Let me know if there are any suggestions and ideas on how to integrate this script with the MMDeploy

Note: 

1. Only tested with "standard" python virtual environment, as I do not use Conda
2. Only tested on linux
3. Only supports TensorRT build at this point.
4. The script should work on both x86_64 and aarch64 (jetson) platforms.


lvhan028 (commented):
First of all, I suggest changing the target branch of this PR to dev-v0.5.0. We would like to make this one a new feature of MMDeploy and announce it when releasing v0.5.0.

build_mmdeploy.sh
```diff
@@ -0,0 +1,346 @@
...
+    # we must build spdlog with -fPIC enabled
+    cmake .. -DCMAKE_POSITION_INDEPENDENT_CODE=ON && make -j
+    sudo make install
+    sudo ldconfig
+  fi
+
+  # cmake check
+  echo_green "Check your cmake version:"
+  cmake --version
+  read -p "Install latest CMake? (>=3.18 is required) (y/n)" -n 1 -r
```
lvhan028, line 71: I think it's better to install `cmake` automatically when its version is less than `v3.14`.
And [CMake](https://cmake.org/download/) already provides the prebuilt binary. For example, 

Linux x86_64 | [cmake-3.23.1-linux-x86_64.sh](https://github.com/Kitware/CMake/releases/download/v3.23.1/cmake-3.23.1-linux-x86_64.sh)
-- | --
  | [cmake-3.23.1-linux-x86_64.tar.gz](https://github.com/Kitware/CMake/releases/download/v3.23.1/cmake-3.23.1-linux-x86_64.tar.gz)
Linux aarch64 | [cmake-3.23.1-linux-aarch64.sh](https://github.com/Kitware/CMake/releases/download/v3.23.1/cmake-3.23.1-linux-aarch64.sh)
  | [cmake-3.23.1-linux-aarch64.tar.gz](https://github.com/Kitware/CMake/releases/download/v3.23.1/cmake-3.23.1-linux-aarch64.tar.gz)

Therefore, instead of building it from the source code, we can download the binary. That would save quite an amount of time.

build_mmdeploy.sh
```diff
@@ -0,0 +1,346 @@
...
+    make -j
+    sudo make install
+    #
+    source ~/.bashrc
+    cmake --version
+  fi
+
+  # gcc-7 check
+  echo_green "Check your gcc version:"
+  gcc --version
```
lvhan028, line 93: Similar to the `cmake` scripts, it's better to update `gcc` automatically when its version is less than 7

lvhan028 (commented):
It seems this script targets NVIDIA devices and Linux OS. So perhaps it can be renamed as `build_linux_nvidia.sh` or something like that.

build_mmdeploy.sh
```diff
@@ -0,0 +1,346 @@
...
+  cmake .. \
+    -DMMDEPLOY_BUILD_SDK=ON \
+    -DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX} \
+    -DCMAKE_CXX_COMPILER=g++ \
+    -Dpplcv_DIR=/usr/local/lib/cmake/ppl \
+    -DMMDEPLOY_TARGET_DEVICES="cuda;cpu" \
+    -DMMDEPLOY_TARGET_BACKENDS=trt \
+    -DMMDEPLOY_CODEBASES=all \
+    -DTENSORRT_DIR=${TENSORRT_DIR} \
+    -DCUDNN_DIR=${CUDNN_DIR}
```
lvhan028, line 285: Please add `-DMMDEPLOY_BUILD_SDK_PYTHON_API=ON` so as to build SDK's python package

grimoire:
The install script of `anaconda` provides some interactive operations to custom the build. For example:

```bash
Anaconda3 will now be installed into this location:
/home/MyHome/anaconda3

  - Press ENTER to confirm the location
  - Press CTRL-C to abort the installation
  - Or specify a different location below

[/home/MyHome/anaconda3] >>> 
```

You can also silence install by adding some other flags. This seems like a good way to custom the build. What do you think @lvhan028 @tehkillerbee

build_mmdeploy.sh
tehkillerbee: @lvhan028 Indeed, the CMake build is slow (I previously used snap install as an alternative but changed it back to building from source). This looks like a cleaner solution so I will change it. It is also a good idea to do this without user interaction.

tehkillerbee:
> The install script of `anaconda` provides some interactive operations to custom the build. For example:
> 
> ```shell
> Anaconda3 will now be installed into this location:
> /home/MyHome/anaconda3
> 
>   - Press ENTER to confirm the location
>   - Press CTRL-C to abort the installation
>   - Or specify a different location below
> 
> [/home/MyHome/anaconda3] >>> 
> ```
> 
> You can also silence install by adding some other flags. This seems like a good way to custom the build. What do you think @lvhan028 @tehkillerbee

This sounds like a good idea for an interactive install procedure. The default options can then be used for an unattended install.

lvhan028:
> The install script of `anaconda` provides some interactive operations to custom the build. For example:
> 
> ```shell
> Anaconda3 will now be installed into this location:
> /home/MyHome/anaconda3
> 
>   - Press ENTER to confirm the location
>   - Press CTRL-C to abort the installation
>   - Or specify a different location below
> 
> [/home/MyHome/anaconda3] >>> 
> ```
> 
> You can also silence install by adding some other flags. This seems like a good way to custom the build. What do you think @lvhan028 @tehkillerbee

Sounds great

build_mmdeploy.sh
tehkillerbee: If an older version of cmake is detected, we should probably purge it and install the latest(?) to the user folder as listed below:

```
# purge existing
sudo apt-get purge cmake
sudo snap remove cmake
#
wget https://github.com/Kitware/CMake/releases/download/v3.23.1/cmake-3.23.1-linux-${ARCH}.sh
chmod +x cmake-3.23.1-linux-${ARCH}.sh
sudo ./cmake-3.23.1-linux-${ARCH}.sh --prefix=/usr --skip-license
```

This is slightly different than the method described in the build documentation for MMDeploy, as we use symlinks to /usr/bin instead of running the install script.

This method should work irrespective of Ubuntu version and we avoid snap packages.

tehkillerbee:
I have updated the PR with most of the comments you both suggested. Whats left is to make the install more interactive and also make it unattended. @lvhan028 @grimoire

grimoire:
Hi, could you drop the unrelated commit (which comes from master branch).

tehkillerbee:
> Hi, could you drop the unrelated commit (which comes from master branch).

I tried to cherry pick the related commits and apply them to the dev v0.5.0 instead. Does it look OK to you?

tehkillerbee:
> The install script of `anaconda` provides some interactive operations to custom the build. For example:
> 
> ```shell
> Anaconda3 will now be installed into this location:
> /home/MyHome/anaconda3
> 
>   - Press ENTER to confirm the location
>   - Press CTRL-C to abort the installation
>   - Or specify a different location below
> 
> [/home/MyHome/anaconda3] >>> 
> ```
> You can also silence install by adding some other flags. This seems like a good way to custom the build. What do you think 

@grimoire Regarding interactive install, should I just attempt to replicate the behaviour used by the anaconda installer or is there a install script I can use as a base?

grimoire:
We do not have a base script, it is ok to replicate the behavior of anaconda or you can design your own pattern.

grimoire:
> I tried to cherry pick the related commits and apply them to the dev v0.5.0 instead. Does it look OK to you?

You can `rebase -i` and drop the commit from master(if they are not required for the build script).

build_linux_nvidia.sh
```diff
@@ -0,0 +1,381 @@
...
+# Directories
+# WORKING_DIR must correspond to script dir, i.e. MMDeploy root
+WORKING_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+PPLCV_DIR=${WORKING_DIR}/ppl.cv
+MMDEPLOY_DIR=${WORKING_DIR}/MMDeploy
+
+#####
+# Versions
+PPLCV_VER="0.6.2"
+MMDEPLOY_VER="0.4.0"
```
grimoire, line 27: Can we read the version from `mmdeploy.version.py`?

build_linux_nvidia.sh
```diff
@@ -0,0 +1,381 @@
...
+  # TODO pack prebuild as tar.gz file
+  #cd ..
+  #tar -zcvf ${WORKING_DIR}/pplcv_${PPLCV_VER}_cuda-${ARCH}-build.tar.gz build/
+}
+
+mmdeploy(){
+  ## mmdeploy SDK
+  cd ${WORKING_DIR}
+  echo_blue "checking out '${MMDEPLOY_DIR}' pkg..."
+  if [ -d "${MMDEPLOY_DIR}" ]; then
```
grimoire, line 285: Since `MMDEPLOY_DIR=${WORKING_DIR}/MMDeploy`, we have to clone it again even when the script is inside the repo dir.

tehkillerbee: @grimoire Yes, you are right. My script was developed with a root working directory, containing MMDeploy. But in our case, the root directory will always be MMDeploy. Therefore we do not need to check out a specific version anymore. I will remove this functionality.

build_linux_nvidia.sh
tehkillerbee: @grimoire The MMDEPLOY_VER is only used for checking out a specific version of MMDeploy. I guess this functionality is not needed anymore and we should just use whatever version that was checked out initially.

build_linux_nvidia.sh
```diff
@@ -0,0 +1,381 @@
...
+    python3 -m venv ${PYTHON_VENV_DIR} --system-site-packages #system site packages to keep trt from system installation
+  fi
+
+  source ${PYTHON_VENV_DIR}/bin/activate
+  python3 get-pip.py
+  pip3 install testresources
+  pip3 install --upgrade setuptools wheel
+
+  # Latest PIL is not compatible with mmcv=1.4.1
+  pip install Pillow==7.0.0
```
grimoire, line 209: pip3 install

build_linux_nvidia.sh
```diff
@@ -0,0 +1,381 @@
...
+    # Install numpy 1.19.4 as newer versions might give "Illegal instruction (core dumped)" on Jetson
+    # TODO Numpy might be installed per default so we should not remove it.
+    pip3 install numpy==1.19.4
+  else
+    pip3 install numpy
+  fi
+  pip3 install opencv-python
+  pip3 install matplotlib
+
+  # pytorch, torchvision, torchaudio
```
grimoire, line 224: Can we make pytorch optional?

build_linux_nvidia.sh
```diff
@@ -0,0 +1,381 @@
...
+  cmake --build . -- -j${PROC_NUM} && sudo cmake --install .
+  sudo ldconfig
+
+  # TODO Pack build output as tar.gz file
+  #cd ..
+  # tar -zcvf ${WORKING_DIR}/mmdeploysdk_${MMDEPLOY_VER}_${ARCH}-build.tar.gz build/
+  # Unpack as tar -zxf mmdeploysdk_*.tar.gz --directory MMDeploy-aarch64
+
+  ## build mmdeploy examples
+  cp -r ${WORKING_DIR}/MMDeploy/demo/csrc ${WORKING_DIR}/MMDeploy/build/example
```
grimoire, line 332: MMDEPLOY_DIR here?

build_linux_nvidia.sh
tehkillerbee: @grimoire Sure, but wont it be required for MMDeploy?

grimoire: We hope users can export the ONNX file on the server and convert it to TensorRT on the board (in the future). Less dependency means less storage usage and faster installation.

tehkillerbee: @grimoire That makes sense. That functionality would also be very valuable for us. For now, we generate the models on the Jetson and copy them to a NAS.

lvhan028:
Hi, @tehkillerbee. I think you can work on dev-v0.5.0 branch since there are lots of unrelated commits from master.

tehkillerbee:
> Hi, @tehkillerbee. I think you can work on dev-v0.5.0 branch since there are lots of unrelated commits from master.

I have rebased it to dev-v0.5.0 now so the unrelated commits should be removed.

lvhan028:
@tehkillerbee Does this script also support silent mode? I am working on circleci #368, and installing in silent mode is a very strong requirement.

tehkillerbee:
@lvhan028 I am going to add silent mode as part of the interactive install. I.e. if the script is executed with a `silent` argument, it will use the default selections for all prompts. 

Will it be OK to use the default selections or will it be necessary to override them somehow?

lvhan028:
> @lvhan028 I am going to add silent mode as part of the interactive install. I.e. if the script is executed with a `silent` argument, it will use the default selections for all prompts.
> 
> Will it be OK to use the default selections or will it be necessary to override them somehow?

Using the default selections is fine. Thanks a lot

build_linux_nvidia.sh
```diff
@@ -0,0 +1,400 @@
...
+    -DMMDEPLOY_CODEBASES=all \
+    -DMMDEPLOY_BUILD_SDK_PYTHON_API=ON \
+    -DTENSORRT_DIR=${TENSORRT_DIR} \
+    -DCUDNN_DIR=${CUDNN_DIR}
+  cmake --build . -- -j${PROC_NUM} && sudo cmake --install .
+  sudo ldconfig
+
+  # generate prebuild and pack into .tar.gz
+  sudo make DESTDIR=./prebuild install
+  tar -zcvf ${WORKING_DIR}/mmdeploysdk_${MMDEPLOY_VER}_${ARCH}-build.tar.gz -C ./prebuild/ .
```
grimoire, line 349: Packaging the library is cool, but I think it can be optional.

lvhan028:
```shell
Build and install 'mmdeploy'...
cat: MMDeploy/mmdeploy/version.py: No such file or directory
./build_linux_nvidia.sh: line 314: /home/PJLAB/lvhan/Documents/projects/open-mmlab/mmdeploy/venv-mmdeploy/bin/activate: No such file or directory
```

tehkillerbee:
> Hi, @tehkillerbee can you provide a build method? Here is what I do in my conda environment:
> 
> ```shell
> git clone --recursive https://github.com/open-mmlab/mmdeploy.git
> cd mmdeploy
> ./build_linux_nvidia.sh mmdeploy
> ```
> 
> But I got failures like,
> 
> ```shell
> Build and install 'mmdeploy'...
> cat: MMDeploy/mmdeploy/version.py: No such file or directory
> ./build_linux_nvidia.sh: line 314: /home/PJLAB/lvhan/Documents/projects/open-mmlab/mmdeploy/venv-mmdeploy/bin/activate: No such file or directory
> ```

I will update the 

> Hi, @tehkillerbee can you provide a build method? Here is what I do in my conda environment:
> 
> ```shell
> git clone --recursive https://github.com/open-mmlab/mmdeploy.git
> cd mmdeploy
> ./build_linux_nvidia.sh mmdeploy
> ```
> 
> But I got failures like,
> 
> ```shell
> Build and install 'mmdeploy'...
> cat: MMDeploy/mmdeploy/version.py: No such file or directory
> ./build_linux_nvidia.sh: line 314: /home/PJLAB/lvhan/Documents/projects/open-mmlab/mmdeploy/venv-mmdeploy/bin/activate: No such file or directory
> ```

I think that is related to the issue mentioned by @grimoire. Initially my script assumed that a MMDeploy folder is found in the root directory. But since the root dir is now MMDeploy, there are some paths that are now incorrect. I will fix it now!

build_linux_nvidia.sh
```diff
@@ -0,0 +1,488 @@
...
+    then
+      return 1
+    elif [[ $REPLY =~ ^[Nn]$ ]]
+    then
+      return 0
+    elif [[ $REPLY =~ ^[Qq]$ ]]
+    then
+      echo_green "Quitting!"
+      exit
+    else
```
grimoire, line 99: How do I choose default option? It give me "invalid argument" if I press enter.

build_linux_nvidia.sh
```diff
@@ -0,0 +1,488 @@
...
+#####
+# Unattended/auto install
+if [[ $appargument2 == "auto" ]]
+then
+  WITH_UNATTENDED=1
+fi
+
+#####
+# Install dependencies in venv?
+prompt_yesno "Install misc. dependencies in the active venv? (Default:${WITH_PYTHON})"
```
grimoire, line 449: It will give me `(Default:1)\n (y/n/q)`.

tehkillerbee: Default option is currently only used when using unattended mode. I will add automatic selection of default option when pressing ENTER, would that work for you?

grimoire: Sounds great.

grimoire:
Good to me. Any advice @lvhan028

build_linux_nvidia.sh
```diff
@@ -0,0 +1,494 @@
...
+    -DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX} \
+    -DCMAKE_CXX_COMPILER=${GCC_COMPILER} \
+    -Dpplcv_DIR=${INSTALL_PREFIX}/lib/cmake/ppl \
+    -DMMDEPLOY_TARGET_DEVICES="cuda;cpu" \
+    -DMMDEPLOY_TARGET_BACKENDS=trt \
+    -DMMDEPLOY_CODEBASES=all \
+    -DMMDEPLOY_BUILD_SDK_PYTHON_API=ON \
+    -DTENSORRT_DIR=${TENSORRT_DIR} \
+    -DCUDNN_DIR=${CUDNN_DIR}
+  cmake --build . -- -j${PROC_NUM} && sudo cmake --install .
```
lvhan028, line 396: `cmake --install .` doesn't work until cmake 3.18
So I suggest using `make install` instead

lvhan028:
Can we put this script to `tools/scripts/`?

build_linux_nvidia.sh
```diff
@@ -0,0 +1,494 @@
...
+    cd spdlog
+    git pull
+    git checkout tags/v1.8.1
+    mkdir build -p && cd build
+    # we must build spdlog with -fPIC enabled
+    cmake .. -DCMAKE_POSITION_INDEPENDENT_CODE=ON && make -j${PROC_NUM}
+    sudo make install
+    sudo ldconfig
+  fi
+  exit
```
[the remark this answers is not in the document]
tehkillerbee: Oops! That was a leftover when I was testing prereqs. I will remove it and update this PR

tehkillerbee:
@lvhan028 Updated PR according to your comments

tools/scripts/build_linux_nvidia.sh
```diff
@@ -0,0 +1,493 @@
...
+BUILD_TYPE="Release"
+ARCH=$(uname -i)
+PROC_NUM=$(nproc)
+# Default GCC
+GCC_COMPILER="g++"
+
+#####
+# Directories
+# WORKING_DIR must correspond to script dir, i.e. MMDeploy root
+WORKING_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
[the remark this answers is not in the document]
tehkillerbee: @lvhan028 Good catch. Yes, the path should be fixed now that we have moved the script into tools/scripts. I have fixed it now

grimoire (approved):
LGTM

PR #381, opened 2022-04-19
[Doc] Improve Jetson tutorial install doc

PeterH0323:
## Motivation

Improve mmdeploy build on Jetson doc: `docs/en/tutorials/how_to_install_mmdeploy_on_jetsons.md`

## Modification

I follow the doc to build mmdeploy on Jetson nano (4G version, Jetpack 4.6.1), and I add some solution or details when I can't get it from the doc.

## Still need to do
- [x] Cmake version of JetPack
- [x] Cmake update guide detail
- [x] Conda install path to `~/archiconda` and show the installation detail

## Checklist

- [x] Pre-commit or other linting tools are used to fix the potential lint issues.
- [x] The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
- [x] If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
- [x] The documentation has been modified accordingly, like docstring or example tutorials.

PR #407, opened 2022-04-25
[BUG] Fix Jetson Nano SDK crash

PeterH0323:
## Motivation

When using Jetson Nano to install mmdeploy and use SDK to predict, it will crash.

6120ce8a2c0497363bece7910adf98d

## Related DOC Imporve
PR #381 

## Modification

`cuda.cmake` in mmdeploy was missing `sm_53`. 

According to website https://en.wikipedia.org/wiki/CUDA , it shows that the minimux version of cuda support `sm_53` is `cuda 7.5`. So I add it in the `cuda.cmake`.

image

After this modifiy, it work.

image

At the same time, I PR to `ppl.cv` [ppl.cv PR#85](https://github.com/openppl-public/ppl.cv/pull/85)

Thx the help from @lzhangzz . 😄


AllentDan (approved):
LGTM

lzhangzz (approved):

PR #426, opened 2022-04-29
[Fix] Fix ci

grimoire:

## Motivation

CUDA ci failed.

## Modification

Fix ci.
The error is caused by a public key error when `apt-get update`

```bash
W: GPG error: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64  InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A4B469963BF863CC
E: The repository 'https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64  InRelease' is not signed.
``` 

which would block all `apt-get install` which include `python-dev`. `setuptools` will not be updated and the installation of `pycocotools` would be failed because of that.

Since this would block ci in every PR, please review it ASAP.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,12 +82,9 @@ jobs:
           apt-get update && apt-get install -y ffmpeg libsm6 libxext6 git ninja-build libglib2.0-0 libxrender-dev python${{matrix.python-version}}-dev
           apt-get clean
           rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


grimoire:
@lvhan028

lvhan028 (approved):

PR #431, opened 2022-05-02
[Fix] ncnn [-Wreturn-type] and improve mutex `if`

PeterH0323:
I using `dev-v0.5.0` for base, If I merge to `master` it will be plenty file changed which is not my commit. So I PR to `v0.5.0`.

## Motivation

When I compile project, I met warning for ncnn : `control reaches end of non-void function [-Wreturn-type]`

## Modification

- Add `return -1` at the end of `Expand::forward` of `csrc/backend_ops/ncnn/ops/expand/expand.cpp`
- Add `return -1` at the end of  `ConstantOfShape::forward` of `csrc/backend_ops/ncnn/ops/constantofshape/constantofshape.cpp`

Also I make the mutex if to `if ... else if ...`

### DIFF
--- a/csrc/backend_ops/ncnn/ops/constantofshape/constantofshape.cpp
+++ b/csrc/backend_ops/ncnn/ops/constantofshape/constantofshape.cpp
@@ -30,7 +30,7 @@ int ConstantOfShape::forward(const Mat& bottom_blob, Mat& top_blob, const Option
     top_blob.fill(val);
     return 0;
   }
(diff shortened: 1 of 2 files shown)


lvhan028 (dismissed):
LGTM

lvhan028 (approved):

grimoire (approved):

hanrui1sensetime (approved):
LGTM.

PR #457, opened 2022-05-10
[Feature] Support reppoints TensorRT

grimoire:

## Motivation

https://github.com/open-mmlab/mmdeploy/issues/439

## Modification

Add reppoints TensorRT support.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/docs/en/benchmark.md
+++ b/docs/en/benchmark.md
@@ -1033,7 +1033,23 @@ Users can directly test the performance through [how_to_evaluate_a_model.md](tut
     <td align="center">40.0</td>
     <td align="center">-</td>
     <td align="center">-</td>
+    <td align="center">-</td>
     <td>$MMDET_DIR/configs/gfl/gfl_r50_fpn_1x_coco.py</td>
+  </tr>
+  <tr>
+    <td align="center">RepPoints</td>
+    <td align="center">Object Detection</td>
+    <td align="center">COCO2017</td>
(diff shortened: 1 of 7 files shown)


AllentDan (commented):
May update Chinese benchmark and supported model list as well.

docs/en/benchmark.md
```diff
@@ -1033,8 +1033,24 @@ Users can directly test the performance through [how_to_evaluate_a_model.md](tut
...
+  <tr>
+    <td align="center">RepPoints</td>
+    <td align="center">Object Detection</td>
+    <td align="center">COCO2017</td>
+    <td align="center">box AP</td>
+    <td align="center">37.0</td>
+    <td align="center">-</td>
+    <td align="center">-</td>
+    <td align="center">36.9</td>
+    <td align="center">20.4</td>
```
RunningLeon, line 1048: is there something wrong with fp16? This is too low compared with fp32.

grimoire: the intermediate result will overflow. There are not much we can do (just like the MMOCR case).

RunningLeon: maybe we could remove fp16 result and add a note.

AllentDan (approved):
LGTM

RunningLeon (approved):
LGTM

PR #479, opened 2022-05-16
If a cuda launch error occurs, verify if cuda device requires top_k t…

### LINKED ISSUE #452: Warning regarding pre_top_k
Hey guys,

When running model generation on Jetson AGX Xavier, we get the following warning:

```
[TensorRT] INFO:   Calibrated batch 41 in 2.60759 seconds.
Warning: pre_top_k need to be reduced for devices with arch 7.2, got pre_top_k=1000
Warning: pre_top_k need to be reduced for devices with arch 7.2, got pre_top_k=1000
```

As I understand from the jetson documentation, this only applies to TX2 and Nano but we get the warning also on AGX. Is this intentional? FYI, the model generation appears to work fine on AGX - even when leavning pre_top_k unchanged or reducing it to 1000, as I did in the above example. 

Do we reduce pre_top_k due to a cuda Compute Capability limitation or a GPU memory limitation?

I seem to remember an issue describing why pre_top_k must be reduced on some Jetson platforms - but I cannot find it.

tehkillerbee:
## Motivation
This PR fixes #452 where an unrelated warning is shown when deploying to Jetson AGX.

## Modification
This PR fixes the issue by making the warning message specific to Jetson Nano & Jetson TX2 and moving the warning into an error check to avoid spamming the log.

### DIFF
--- a/csrc/backend_ops/tensorrt/batched_nms/allClassNMS.cu
+++ b/csrc/backend_ops/tensorrt/batched_nms/allClassNMS.cu
@@ -212,7 +212,17 @@ pluginStatus_t allClassNMS_gpu(cudaStream_t stream, const int num, const int num
       (T_BBOX *)bbox_data, (T_SCORE *)beforeNMS_scores, (int *)beforeNMS_index_array,
       (T_SCORE *)afterNMS_scores, (int *)afterNMS_index_array, flipXY);
 
-  CSC(cudaGetLastError(), STATUS_FAILURE);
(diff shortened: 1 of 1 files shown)


AllentDan:
Hi, @tehkillerbee, you may fix the lint first please.

tehkillerbee:
@AllentDan Sorry, I thought it was already fixed. It should be OK now!

AllentDan (approved):
LGTM

lzhangzz (approved):
LGTM

PR #480, opened 2022-05-17
fix mmseg twice resize

RunningLeon:
## Motivation

Unnecessary resize in mmseg encode-decoder head: https://github.com/open-mmlab/mmsegmentation/issues/1529

## Modification

- [x] Remove second unused resize.
- [x] Make sure output of 4d tensor in a nice way.
- [x] Update unit test.

## BC-breaking (Optional)

None

## Use cases (Optional)

### DIFF
--- a/mmdeploy/codebase/mmseg/models/segmentors/encoder_decoder.py
+++ b/mmdeploy/codebase/mmseg/models/segmentors/encoder_decoder.py
@@ -1,9 +1,7 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import torch.nn.functional as F
-from mmseg.ops import resize
 
 from mmdeploy.core import FUNCTION_REWRITER
-from mmdeploy.utils import is_dynamic_shape
 
 
 @FUNCTION_REWRITER.register_rewriter(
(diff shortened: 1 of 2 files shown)


mmdeploy/codebase/mmseg/models/segmentors/encoder_decoder.py
```diff
@@ -25,16 +23,7 @@ def encoder_decoder__simple_test(ctx, self, img, img_meta, **kwargs):
...
-        mode='bilinear',
-        align_corners=self.align_corners)
     seg_logit = F.softmax(seg_logit, dim=1)
-    seg_pred = seg_logit.argmax(dim=1)
-    # our inference backend only support 4D output
-    shape = seg_pred.shape
-    if not is_dynamic_shape(ctx.cfg):
-        shape = [int(_) for _ in shape]
-    seg_pred = seg_pred.view(shape[0], 1, shape[1], shape[2])
+    # keep 4D output required by backends such as TensorRT
```
grimoire, line 27: It is not required by any backends.
We do a nearest resize outside the model in `segmentation_model` as the postprocess, which require 4d input.

grimoire (commented):
Moving the `resize` out of the model is a good optimization.
But the accuracy of the model might be affected. should we update the benchmark?

mmdeploy/codebase/mmseg/models/segmentors/encoder_decoder.py
RunningLeon: It seems that trt requires 4d output as tested before when supporting deployment feature in mmseg

grimoire: It is caused by postprocessing even in mmseg. We can just remove this comment.

grimoire (approved):

AllentDan (approved):
LGTM

PR #483, opened 2022-05-17
[Fix] set optional arg a default value

AllentDan:
## Motivation
`--work-dir` is an optional arg of deploy.py while missing the default value.

### DIFF
--- a/tools/deploy.py
+++ b/tools/deploy.py
@@ -1,6 +1,7 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import argparse
 import logging
+import os
 import os.path as osp
 from functools import partial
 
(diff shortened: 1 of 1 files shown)


tools/deploy.py
```diff
@@ -26,7 +27,8 @@ def parse_args():
     parser.add_argument('img', help='image used to convert model model')
     parser.add_argument(
         '--test-img', default=None, help='image used to test model')
-    parser.add_argument('--work-dir', help='the dir to save logs and models')
+    parser.add_argument(
+        '--work-dir', default=None, help='the dir to save logs and models')
```
grimoire, line 31: Can we set `os.getcwd()` as default here?

grimoire (approved):

tpoisonooo:
LGTM.

tpoisonooo (approved):

PR #521, opened 2022-05-25
Profiler

irexyc:
## Motivation

profiler

## Modification


csrc/graph/pipeline.cpp
```diff
@@ -100,7 +101,15 @@ Sender<Value> Pipeline::Process(Sender<Value> args) {
   State state(use_count_, std::move(args));
   for (size_t i = 0; i < nodes_.size(); ++i) {
     auto input = state.Collect(input_coords_[i]);
+    input = Then(input, [name = nodes_[i]->name()](Value&& input) {
```
lzhangzz, line 104: type earsed senders may not be copyable, always use moving to pass senders unless copying is intended.

PR #542, opened 2022-05-31
[RegressionTest] Add mmrotate regression test

grimoire:

## Motivation

Add regression test config for mmrotate. Rotated Retinanet only.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/tests/regression/mmrotate.yml
+++ b/tests/regression/mmrotate.yml
@@ -1,0 +1,50 @@
+globals:
+  codebase_dir: ../mmrotate
+  checkpoint_force_download: False
+  images:
+    img_demo: &img_demo ../mmrotate/demo/demo.jpg
+    img_dota_demo: &img_dota_demo ../mmrotate/demo/dota_demo.jpg
+  metric_info: &metric_info
+    mAP: # named after metafile.Results.Metrics
+      eval_name: mAP # test.py --metrics args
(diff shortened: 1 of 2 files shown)


tests/regression/mmrotate.yml
```diff
@@ -0,0 +1,50 @@
+globals:
```
RunningLeon, line 1: should we update mmrotate support in `tools/regression_test.py` accordingly?

RunningLeon, line 1: should we update mmrotate support in `tools/regression_test.py` accordingly?

grimoire: I am not so sure what should I add.
Actually, I can not test the accuracy since the dataset require update the result to the server.

lvhan028: I think we should support it. We can test model conversion at least. 
What are the concerns?

RunningLeon: Agree. Conversion could be tested.

RunningLeon (approved):
LGTM

tests/regression/mmrotate.yml
RunningLeon: All conversions tested success on my side.

lvhan028 (approved):

PR #574, opened 2022-06-08
bump version to v0.5.0

lvhan028:


.github/scripts/doc_link_checker.py
```diff
@@ -0,0 +1,77 @@
+# Copyright (c) MegFlow. All rights reserved.
```
AllentDan, line 1: may check again all copyrights are right

lzhangzz (approved):

RunningLeon (approved):
LGTM

PR #576, opened 2022-06-08
Fix mmpose regression test failed on mmpose v0.26.0

irexyc:
## Motivation

MMPose v0.26.0 move the center and scale calculation to the transform. 

This pr make regression_test.py works on mmpose v0.26.0 and v0.25.0.


configs/mmpose/pose-detection_sdk_static-256x192.py
```diff
@@ -0,0 +1,15 @@
+_base_ = ['./pose-detection_static.py', '../_base_/backends/sdk.py']
```
RunningLeon, line 1: should we remove old sdk config `configs/mmpose/pose-detection_sdk_static.py`? Is this new sdk config only for mmpose>=v0.26.0?

irexyc: Coco keypoint dataloader in mmpose v0.25.0 will calculate *center* and *scale* with dataset annotation information. When using sdk backend, the annotation information will lost. The main purpose of new sdk config with size is bring the annotation information to the end2end sdk model to calculate the original *center* and *scale* passed to coco evaluation tools. (may not used in coco, but not sure about other dataset)

RunningLeon (approved):
LGTM

lzhangzz (approved):

PR #594, opened 2022-06-13
[Enhancement] Adjust `csrc` structure

lzhangzz:


irexyc:
1. can't compile demo on windows.
2. doesn't modify the demo include header like
```cpp
#include "pose_detector.h" // include "mmdeploy/c/pose_detector.h"
```

csrc/mmdeploy/apis/c/executor.cpp
```diff
@@ -1,11 +1,11 @@
 // Copyright (c) OpenMMLab. All rights reserved.
 
-#include "apis/c/executor.h"
+#include "mmdeploy/apis/c/executor.h"
 
 #include "common.h"
 #include "common_internal.h"
-#include "execution/when_all_value.h"
 #include "executor_internal.h"
+#include "mmdeploy/execution/when_all_value.h"
```
lvhan028, line 8: Can we include the header file in the same pattern, e.g., the relative path to `${MMDEPLOY_ROOT_DIR}/csrc` ?

lvhan028 (approved):

PR #603, opened 2022-06-15
add dim param for `Tensor::Squeeze`

lzhangzz:


irexyc (approved):

hanrui1sensetime (approved):
LGTM but still has bugs in ocr c demo, it is independent to this pr. I will fix these bugs in an another pr.

PR #609, opened 2022-06-16
[WIP] [experimental feature] Add option to fuse transform.

irexyc:
## Motivation

Use Elena to fuse transform.

## Modification

add some dummy compute node to skip compute and write transform args to trans_info.


lvhan028:
`json.hpp` in `fuse_kernel` is not necessary because mmdeploy already has `third_party/json/json.hpp`. Please use it instead

irexyc:
> `json.hpp` in `fuse_kernel` is not necessary because mmdeploy already has `third_party/json/json.hpp`. Please use it instead

When use fuse function, I think it's better to extract runtime args in the fuse lib rather than mmdeploy. So, this folder is to remove later after introduce elena, and the code in this folder should compile together with source.c that elena genereates.

csrc/mmdeploy/preprocess/elena/elena_registry.h
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) OpenMMLab. All rights reserved.
+
```
lzhangzz, line 2: use core/registry.h instead of inventing a new registry type

csrc/mmdeploy/core/tracer.h
```diff
@@ -0,0 +1,96 @@
...
+                DataType dtype);
+
+  void TraceNorm(const std::vector<float> &mean, const std::vector<float> &std, bool to_rgb,
+                 DataType dtype);
+
+  void TraceCrop(const std::vector<int> &tlbr, const std::vector<int> &size, DataType dtype);
+
+  void TraceDFB(bool to_float, DataType dtype);
+
+  void TraceIm2Tensor(DataType dtype);
```
lzhangzz, line 80: drop redundant "Trace"s and use function names consistent with transforms.

PR #616, opened 2022-06-18
[Fix] Add `pycuda` in requirement

PeterH0323:
## Motivation

Using mmdeploy 0.5.0 convert model to trt, got error: `no model found pycuda`

## Modification

Add `pycuda==2021.1` in `requirements/runtime.txt`

## Note

But this may cause error for those machines which CPU only.

## Checklist


PeterH0323:
I check the doc, It already have some note about pycuda, So I close this PR.

Issue #613, opened 2022-06-17
A bug in c# sdk demo maybe


jrChen2022a:
When running mmdeploy/demo/csharp/Demo.sln in VS2019 in windows11，an error, 
'System.AccessViolationException:“Attempted to read or write protected memory. This is often an indication that other memory is corrupt.” '
, sometimes occurrs in line 52, 
'List<DetectorOutput> output = handle.Apply(mats);'.
Like:
<img width="699" alt="mmexport1655459458335" src="https://user-images.githubusercontent.com/100680047/174274526-0f00d488-472a-4200-b4da-c46b91fec527.png">

jrChen2022a:
sorry, wrong picture

caj-github:
same problem in win10
[2022-06-19 10:45:49.924] [mmdeploy] [warning] [bulk.h:39] fallback Bulk implementation
[2022-06-19 10:45:49.933] [mmdeploy] [warning] [bulk.h:39] fallback Bulk implementation
[2022-06-19 10:45:50.574] [mmdeploy] [warning] [bulk.h:39] fallback Bulk implementation
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at MMDeploy.NativeMethods.mmdeploy_detector_release_result(MMDeploy.CDetect*, Int32*, Int32)
   at MMDeploy.NativeMethods.mmdeploy_detector_release_result(MMDeploy.CDetect*, Int32*, Int32)
   at MMDeploy.Detector.ReleaseResult(MMDeploy.CDetect*, Int32*, Int32)
   at MMDeploy.Detector.Apply(MMDeploy.Mat[])
   at object_detection.Program.Main(System.String[])

caj-github:
i changed source code, it works now

first change in mmdeploy\csrc\mmdeploy\backend_ops\tensorrt\multi_level_roi_align.cpp\trt_multi_level_roi_align.cpp

![image](https://user-images.githubusercontent.com/55951697/174470466-656b86cf-e90e-468d-8428-cc13092775be.png)

second change in mmdeploy\csrc\mmdeploy\apis\csharp\MMDeploy\APIs\Detector.cs
![image](https://user-images.githubusercontent.com/55951697/174470592-8a015df7-a563-410f-9697-c71d4633e05a.png)

irexyc:
The `total` args passed to ReleaseResult is wrong. 
[...]

PR #620, opened 2022-06-20
[Fix] fix csharp api detector release result

irexyc:
## Motivation

to address https://github.com/open-mmlab/mmdeploy/issues/613

Fix wrong args in mmdeploy_detector_release_result


csrc/mmdeploy/apis/csharp/MMDeploy/APIs/Detector.cs
```diff
@@ -205,15 +205,14 @@ public List<DetectorOutput> Apply(Mat[] mats)
 
         private unsafe void FormatResult(int matCount, int* resultCount, CDetect* results, ref List<DetectorOutput> output, out int total)
         {
-            total = 0;
+            total = matCount;
```
lvhan028, line 208: So `total` means the number of images?

lzhangzz (approved):

lvhan028 (approved):

Issue #631, opened 2022-06-22
Any support on cuda11.x version?


XWHtorrentx:
I notice that dockerfile build-up is based on cuda10.2. Is it possible to build with later version of torch, mmcv, onnxruntime and cuda11.x? 

If the answer is yes please be specific about versions of prerequisites, thanks!

AllentDan:
You may pass the `CUDA=11.x` arg in to the dockerfile when building docker image.

XWHtorrentx:
> 

I have tried with CUDA=11.1 (what is supposed to be the most common choice). However conda won't install pytorch LTS with cuda support.
I modified the command with official one : `conda install pytorch torchvision torchaudio cudatoolkit=11.1 -c pytorch-lts -c nvidia`, and it turns out that I cannot have pytorch installed with cuda support either.

PR #636, opened 2022-06-23
Add dockerfile supporting cuda11 build-up

XWHtorrentx:
add a dockerfile for cuda11 build-up

I modified the version of torch, torchvision, cudatoolkit, and the command line to install torch, mmcv.

### DIFF
--- a/docker/GPU/Dockerfilecu11
+++ b/docker/GPU/Dockerfilecu11
@@ -1,0 +1,96 @@
+FROM nvcr.io/nvidia/tensorrt:21.04-py3
+
+ARG CUDA=11.1
+ARG PYTHON_VERSION=3.8
+ARG TORCH_VERSION=1.8.2
+ARG TORCHVISION_VERSION=0.9.2
+ARG ONNXRUNTIME_VERSION=1.8.1
+ARG MMCV_VERSION=1.4.0
+ARG PPLCV_VERSION=0.6.2
+ENV FORCE_CUDA="1"
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+### change the system source for installing libs
+ARG USE_SRC_INSIDE=false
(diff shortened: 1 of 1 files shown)


lvhan028:
It is not much different from `docker/GPU/Dockerfile`. So you can change it directly without making a new file.
But I am curious about the PyTorch installation command you are using. 
From its official guide, it suggests
```shell
conda install pytorch==${TORCH_VERSION} torchvision==${TORCHVISION_VERSION} cudatoolkit=${CUDA} -c pytorch -c conda-forge
```
So is `-c conda-forge` able to resolve issue #631? 
If it is, I think we'd better use the official one.

XWHtorrentx:
> It is not much different from `docker/GPU/Dockerfile`. So you can change it directly without making a new file. But I am curious about the PyTorch installation command you are using. From its official guide, it suggests
> 
> ```shell
> conda install pytorch==${TORCH_VERSION} torchvision==${TORCHVISION_VERSION} cudatoolkit=${CUDA} -c pytorch -c conda-forge
> ```
> 
> So is `-c conda-forge` able to resolve issue #631? If it is, I think we'd better use the official one.

I tried `-c conda-forge`, it ended up with a cpu build-up installed, sometimes it had issue of environment resolving. 
Actually the command I used is the official one for installing pytorch LTS on Linux. It worked well.

lvhan028:
I find that PyTorch `v1.8.2` is not in https://pytorch.org/get-started/previous-versions/.
I tried to use `-c conda-forge` to install it and got the following errors indeed.
image

I am not sure if `-c pytorch-lts nvidia` will bring other issues for non-lts PyTorch.

So from the perspective of compatibility, I think we can add an argument to indicate `lts`. Based on this argument and CUDA toolkit version, we can use different ways to install PyTorch.

PR #667, opened 2022-06-28
[FIX] fix dump info of mmpose if input_shape is not given in deploy_cfg

irexyc:
## Motivation

When export mmpose model with --dump-info, if deploy_cfg doesn't contains input_shape, we should consider data_cfg['image_size'] in model config.


lzhangzz (approved):
LGTM

lvhan028 (approved):

PR #715, opened 2022-07-07
fix wrong exit code in pipeline_manager

RunningLeon:
## Motivation

fix wrong exit code in pipeline_manager. Tools like regression_test.py need a nonzero exit code if the program exits wrongly.

## Modification

Set exit code to -1 in pipeline_manager

## BC-breaking (Optional)

None

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdeploy/apis/core/pipeline_manager.py
+++ b/mmdeploy/apis/core/pipeline_manager.py
@@ -79,7 +79,7 @@ class PipelineCaller:
         if call_id not in self._mp_dict:
             get_root_logger().error(
                 f'`{self._func_name}` with Call id: {call_id} failed. exit.')
-            exit(-1)
+            exit(1)
         ret = self._mp_dict[call_id]
         self._mp_dict.pop(call_id)
         return ret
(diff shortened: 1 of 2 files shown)


mmdeploy/apis/core/pipeline_manager.py
```diff
@@ -79,7 +79,7 @@ def pop_mp_output(self, call_id: int = None) -> Any:
         if call_id not in self._mp_dict:
             get_root_logger().error(
                 f'`{self._func_name}` with Call id: {call_id} failed. exit.')
-            exit()
+            exit(-1)
```
grimoire, line 82: -1 is not a vaild exit status

https://www.baeldung.com/linux/status-codes

AllentDan (approved):
LGTM

mmdeploy/backend/pplnn/utils.py
```diff
@@ -37,14 +37,14 @@ def register_engines(device_id: int,
...
-            sys.exit(-1)
+            sys.exit(1)
 
         if disable_avx512:
             status = x86_engine.Configure(pplnn.X86_CONF_DISABLE_AVX512)
             if status != pplcommon.RC_SUCCESS:
                 logger.error('x86 engine Configure() failed: ' +
                              pplcommon.GetRetCodeStr(status))
-                sys.exit(-1)
+                sys.exit(1)
```
AllentDan, line 47: seems `1` is valid but `-1` is from [pplnn](https://github.com/openppl-public/ppl.nn/blob/01314256e227625757a7ad7aa95fb9c28dc69dc5/tools/pplnn.py#L200)

grimoire (approved):

Issue #708, opened 2022-07-06
error ocured when convert onnx to tensorrt on window10


centerzong:
Thanks for your bug report. We appreciate it a lot.

**Checklist**

1. I have searched related issues but cannot get the expected help.
2. I have read the [FAQ documentation](https://github.com/open-mmlab/mmdeploy/blob/master/docs/en/faq.md) but cannot get the expected help.
3. The bug has not been fixed in the latest version.

**Describe the bug**

**Reproduction**

1. What command or script did you run?

```none
A placeholder for the command.
```

2. Did you make any modifications on the code or config? Did you understand what you have modified?

**Environment**

1. Please run `python tools/check_env.py` to collect necessary environment information and paste it here.
WARNING: The shape inference of mmdeploy::TRTBatchedNMS type is missing, so it may result in wrong shape inference for the exported graph. Please consider adding it in symbolic function.
2022-07-06 10:00:02,645 - mmdeploy - INFO - Finish pipeline mmdeploy.apis.pytorch2onnx.torch2onnx
2022-07-06 10:00:23,286 - mmdeploy - INFO - Start pipeline mmdeploy.backend.tensorrt.onnx2tensorrt.onnx2tensorrt in subprocess
2022-07-06 10:00:24,096 - mmdeploy - INFO - Successfully loaded tensorrt plugins from d:\workspace\aiwork\ai-desktop\mmdeploy\mmdeploy\lib\mmdeploy_tensorrt_ops.dll
[07/06/2022-10:00:29] [TRT] [I] [MemUsageChange] Init CUDA: CPU +525, GPU +0, now: CPU 4637, GPU 1273 (MiB)
[07/06/2022-10:00:32] [TRT] [I] [MemUsageSnapshot] Begin constructing builder kernel library: CPU 4683 MiB, GPU 1273 MiB
[07/06/2022-10:00:32] [TRT] [I] [MemUsageSnapshot] End constructing builder kernel library: CPU 4866 MiB, GPU 1317 MiB
[07/06/2022-10:00:33] [TRT] [W] onnx2trt_utils.cpp:366: Your ONNX model has been generated with INT64 weights, while TensorRT does not natively support INT64. Attempting to cast down to INT32.
[07/06/2022-10:00:33] [TRT] [W] onnx2trt_utils.cpp:392: One or more weights outside the range of INT32 was clamped
[...]

PR #751, opened 2022-07-14
[Fix][P0] Fix for torch1.12

grimoire:
1. PyTorch will inspect the rewritten symbolic function(which has been replaced with a ContextCaller)

https://github.com/pytorch/pytorch/blob/67ece03c8cd632cce9523cd96efde6f2d1cc8121/torch/onnx/utils.py#L1333

2. PyTorch add a new pass `_jit_pass_onnx_deduplicate_initializers`, which will turn `k` in `topk` from `initializer` to `Identity`. Which broken the TensorRT export.

https://github.com/pytorch/pytorch/blob/67ece03c8cd632cce9523cd96efde6f2d1cc8121/torch/onnx/utils.py#L1101

### DIFF
--- a/mmdeploy/core/rewriters/rewriter_utils.py
+++ b/mmdeploy/core/rewriters/rewriter_utils.py
@@ -2,6 +2,7 @@
 import inspect
 import warnings
 from abc import ABCMeta, abstractmethod
+from functools import wraps
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import mmdeploy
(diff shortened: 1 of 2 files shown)


mmdeploy/core/rewriters/rewriter_utils.py
```diff
@@ -352,6 +353,11 @@ def __init__(self, func: Callable, origin_func: Callable, cfg: Dict,
         self.func = func
         self.origin_func = origin_func
         self.cfg = cfg
+        if origin_func is not None:
```
RunningLeon, line 356: may add comment for explanation

RunningLeon (approved):
LGTM

lvhan028:
Hi, @irexyc Please verify if this PR can fix https://github.com/open-mmlab/mmdeploy/issues/708

irexyc (approved):

PR #754, opened 2022-07-15
add version control

AllentDan:
Add version control to padding rewriting.


grimoire:
Why do we need this version check?

AllentDan:
> Why do we need this version check?

For torch 1.10-, the arguments of the function are different.

grimoire (approved):

RunningLeon (approved):

PR #706, opened 2022-07-05
Upgrade Dockerfile to use TensorRT==8.2.4.2

nijkah:
## Motivation
Currently, many documents seem to suppose that the user installed TRT >= 8.*.
But, Dockerfile does not follow it, therefore related issues happen when using a given Dockerfile.

## Modification
- Upgrade the base image to use TensorRT 8.2.4.2. See [release note](https://docs.nvidia.com/deeplearning/tensorrt/container-release-notes/rel_22-04.html#rel_22-04).
  - Fix wrong environment variable for `LD_LIBRARY_PATH` (the image does not include `/usr/local/cuda-11.3/`).
- Upgrade torch & torchvision
- Upgrade MMCV

## BC-breaking (Optional)

Does the modification introduce changes that break the backward-compatibility of the downstream repositories?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
3. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
4. The documentation has been modified accordingly, like docstring or example tutorials.

Issue #460, opened 2022-05-10
TensorRT deploy error after upgrading version


austinmw:
Hi, I attempted to upgrade the GPU Dockerfile to use TensorRT `21.08` in order to make it compatible with my Triton inference container version. Before upgrading I was able to successfully export my yolox model to an `end2end.engine` file, but after upgrading I face the following error:

> [2022-05-10 18:38:06.082] [mmdeploy] [info] [model.cpp:95] Register 'DirectoryModel'
> 2022-05-10 18:38:06,147 - mmdeploy - INFO - torch2onnx start.
> [2022-05-10 18:38:07.800] [mmdeploy] [info] [model.cpp:95] Register 'DirectoryModel'
> 2022-05-10:18:38:08,matplotlib.font_manager INFO     [font_manager.py:1443] generated new fontManager
> load checkpoint from local path: /volume_share/checkpoint.pth
> The model and loaded state dict do not match exactly
> 
> unexpected key in source state_dict: ema_backbone_stem_conv_conv_weight, ema_backbone_stem_conv_bn_weight, 
> ...SHORTENED BY ME TO REDUCE LOGS
> 
> 2022-05-10 18:38:13,432 - mmdeploy - WARNING - DeprecationWarning: get_onnx_config will be deprecated in the future. 
> 2022-05-10:18:38:13,mmdeploy WARNING  [utils.py:91] DeprecationWarning: get_onnx_config will be deprecated in the future. 
> /root/workspace/mmdeploy/mmdeploy/core/optimizers/function_marker.py:158: TracerWarning: Converting a tensor to a Python integer might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
>   ys_shape = tuple(int(s) for s in ys.shape)
> /opt/conda/lib/python3.8/site-packages/torch/functional.py:568: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at  /opt/conda/conda-bld/pytorch_1646755903507/work/aten/src/ATen/native/TensorShape.cpp:2228.)
>   return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]
[...]

PR #758, opened 2022-07-18
fix yolox point_generator

AllentDan:
Fix shape inference error raised by new versions of TensorRT. The bug is triggered not only in the docker like issue #460 and the PR #706, but also in the local Ubuntu with TensorRT 8.4.1.5.

### DIFF
--- a/mmdeploy/codebase/mmdet/core/point_generator.py
+++ b/mmdeploy/codebase/mmdet/core/point_generator.py
@@ -1,0 +1,66 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import torch
+
+from mmdeploy.core import FUNCTION_REWRITER
+from mmdeploy.utils.constants import Backend
+
+
+@FUNCTION_REWRITER.register_rewriter(
+    func_name='mmdet.core.anchor.MlvlPointGenerator.single_level_grid_priors',
+    backend=Backend.TENSORRT.value)
+def mlvl_point_generator__single_level_grid_priors__tensorrt(
+        ctx,
+        self,
+        featmap_size,
+        level_idx,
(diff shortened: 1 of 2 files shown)


mmdeploy/codebase/mmdet/core/point_generator.py
```diff
@@ -0,0 +1,66 @@
...
+        ctx,
+        self,
+        featmap_size,
+        level_idx,
+        dtype=torch.float32,
+        device='cuda',
+        with_stride=False):
+    """Rewrite `single_level_grid_priors` of `MlvlPointGenerator` as
+    onnx2tensorrt raise the error of shape inference for YOLOX with some
+    versions ofTensorRT.
```
grimoire, line 21: typo.

mmdeploy/codebase/mmdet/core/point_generator.py
```diff
@@ -0,0 +1,66 @@
...
+
+    shift_y = (torch.arange(0, feat_h, device=device) + self.offset) * stride_h
+    # keep featmap_size as Tensor instead of int, so that we
+    # can convert to ONNX correctly
+    shift_y = shift_y.to(dtype)
+    shift_xx, shift_yy = self._meshgrid(shift_x, shift_y)
+    if not with_stride:
+        shifts = torch.stack([shift_xx, shift_yy], dim=-1)
+    else:
+        # use `shift_x.shape[0] * shift_y.shape[0]` instead of
```
grimoire, line 58: Can we use `feat_h` and `feat_w`? They might be able to further ease the graph.

lvhan028 (approved):

grimoire (approved):

PR #785, opened 2022-07-21
[Fix] fix check env

AllentDan:
Use `is_custom_ops_available` instead of `is_available` for check_env.py.

### DIFF
--- a/tools/check_env.py
+++ b/tools/check_env.py
@@ -25,15 +25,15 @@ def check_backend():
     import mmdeploy.apis.onnxruntime as ort_apis
     logger = get_root_logger()
     logger.info(f'onnxruntime: {ort_version}\tops_is_avaliable : '
-                f'{ort_apis.is_available()}')
+                f'{ort_apis.is_custom_ops_available()}')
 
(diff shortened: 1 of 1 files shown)


tools/check_env.py
```diff
@@ -25,15 +25,15 @@ def check_backend():
     import mmdeploy.apis.onnxruntime as ort_apis
     logger = get_root_logger()
     logger.info(f'onnxruntime: {ort_version}\tops_is_avaliable : '
-                f'{ort_apis.is_available()}')
+                f'{ort_apis.is_custom_ops_available()}')
 
     import mmdeploy.apis.tensorrt as trt_apis
     logger.info(f'tensorrt: {trt_version}\tops_is_avaliable : '
-                f'{trt_apis.is_available()}')
+                f'{trt_apis.is_custom_ops_available()}')
```
lvhan028, line 32: So, we don't check if custom ops have been built? Why?

grimoire (approved):

tools/check_env.py
lvhan028: My bad.

lvhan028 (approved):

Issue #795, opened 2022-07-22
TensorRT int8 model test failed when batch size>1


DarrenCheung:
Network: yolov3
Config file: configs/yolo/yolov3_d53_mstrain-416_273e_coco.py

We use "configs/mmdet/detection/detection_tensorrt-int8_dynamic-64x64-608x608.py" within mmdeploy for TRT int8 model quant, no errors occurs until when run ./mmdeploy/tools/test.py for model test.

Errors:
```
2022-07-22 07:34:54,879 - mmdeploy - INFO - Successfully loaded tensorrt plugins from /opt/conda/lib/python3.8/site-packages/mmdeploy/lib/libmmdeploy_tensorrt_ops.so
2022-07-22 07:34:54,879 - mmdeploy - INFO - Successfully loaded tensorrt plugins from /opt/conda/lib/python3.8/site-packages/mmdeploy/lib/libmmdeploy_tensorrt_ops.so
NOTE! Installing ujson may make loading annotations faster.
loading annotations into memory...
Done (t=0.61s)
creating index...
index created!
[                                                  ] 0/5000, elapsed: 0s, ETA:Traceback (most recent call last):
  File "tools/test.py", line 138, in <module>
    main()
  File "tools/test.py", line 130, in main
    outputs = task_processor.single_gpu_test(model, data_loader, args.show,
  File "/opt/conda/lib/python3.8/site-packages/mmdeploy/codebase/base/task.py", line 138, in single_gpu_test
    return self.codebase_class.single_gpu_test(model, data_loader, show,
  File "/opt/conda/lib/python3.8/site-packages/mmdeploy/codebase/mmdet/deploy/mmdetection.py", line 142, in single_gpu_test
    outputs = single_gpu_test(model, data_loader, show, out_dir, **kwargs)
  File "/workspace/algorithm/tensorrt_infer/yolov3/mmdetection/mmdet/apis/test.py", line 26, in single_gpu_test
    for i, data in enumerate(data_loader):
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 521, in __next__
    data = self._next_data()
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1203, in _next_data
    return self._process_data(data)
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1229, in _process_data
    data.reraise()
[...]

PR #829, opened 2022-07-28
support test with multi batch

RunningLeon:
## Motivation
Add `--batch-size` argument to support run `tools/test.py` with multi batch inference.

## Modification

1. Add `--batch-size` argument to support run `tools/test.py` with multi batch inference.
2. Fix mmdet batch inference in #795

## BC-breaking (Optional)
None
## Use cases (Optional)

```
python tools/test.py \
configs/mmcls/classification_tensorrt_dynamic-224x224-224x224.py \
../mmclassification/configs/resnet/resnet18_8xb32_in1k.py \ 
--model ./work-dirs/mmcls/resnet/trt/end2end.engine \
--device cuda \
--batch-size 64 
```

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -73,6 +73,12 @@ def parse_args():
         help='the interval between each log, require setting '
         'speed-test first',
         default=100)
+    parser.add_argument(
+        '--batch-size',
+        type=int,
+        default=1,
+        help='the batch size for test, would override `samples_per_gpu`'
+        'in  data config.')
 
     args = parser.parse_args()
     return args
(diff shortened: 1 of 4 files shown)


tools/test.py
```diff
@@ -97,9 +103,11 @@ def main():
     # prepare the dataset loader
     dataset_type = 'test'
     dataset = task_processor.build_dataset(model_cfg, dataset_type)
+    # override samples_per_gpu that used for training
+    model_cfg.data['samples_per_gpu'] = args.batch_size
     data_loader = task_processor.build_dataloader(
         dataset,
-        samples_per_gpu=1,
+        samples_per_gpu=model_cfg.data.samples_per_gpu,
```
AllentDan, line 110: Shall we update the latency performance if we use batch_size >1?

grimoire (approved):

tools/test.py
RunningLeon: Maybe not for now. not sure if all models support exporting with batch_size >1.

AllentDan: How about adding a TODO here? Besides, if not for all the models, shall we add limitations to the use of `--batch-size`?

RunningLeon: Adding a note in the description is Ok.

AllentDan (approved):

Issue #659, opened 2022-06-27
Deploy mmpose with webcam_demo


baigwadood:
Thanks for great repository
How can i Deploy mmpose with webcam_demo?

lvhan028:
@antileaf is working on it now.

baigwadood:
@AntiLeaf do mention me once it will be done

baigwadood:
@AntiLeaf any update?

AntiLeaf:
> @AntiLeaf any update?

Sorry for my absence due to personal reasons before. I am working on it now, and I guess it will be done soon.

Once it's done I will mention you. Thanks for your follow. @baigwadood

PR #853, opened 2022-08-02
[feat] add basic webcam demo (#659)

AntiLeaf:
## Motivation

In order to get started more quickly, a webcam demo which is easy-to-use is needed. Also solving issue #659.

## Modification

Added `demo/python/webcam_demo.py`, `demo/python/webcam_demo/__init__.py` and `demo/python/webcam_demo/misc.py`.

## BC-breaking (Optional)

There is no BC-breaking.

## Use cases (Optional)

The webcam demo can be used to show how to take use of the APIs of MMDeploy. This is a basic version, and more features may be added soon.

## Checklist

- [x] Pre-commit or other linting tools are used to fix the potential lint issues.
- [x] The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
- [x] If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
- [ ] The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/demo/python/webcam_demo.py
+++ b/demo/python/webcam_demo.py
@@ -1,0 +1,62 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import logging
+from argparse import ArgumentParser
+
+import webcam_demo
+
+# from webcam_demo.misc import is_image, is_video
+
+
+def parse_args():
+    parser = ArgumentParser('')
+
+    parser.add_argument('model_path')
+
+    parser.add_argument('--device', type=str, default='cuda')
+
+    parser.add_argument('--camera', type=str, default='0')
+
+    parser.add_argument('--debug', action='store_true')
+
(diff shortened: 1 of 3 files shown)


lvhan028:
Please merge master to resolve readthedocs build error

irexyc:
I think the file structure is not appropriate. Maybe we can add a folder in demo/python and move the related files to it.

It seems like it only support specific models, it should explain in readme.

The strategy of check_model is not good, we can read the meta info from detail.json and deploy.json.

An actual run command is needed. and it's better to have a demo picture.

To lower the bar of running the demo, the user guide should be more detailed. And it's better to have a link of pre-converted models

ChenZhenGui:
hello, I want to ask how to use the models.py while my model is not the DEEPPOSE_RESNET?

PR #890, opened 2022-08-11
[Feature] TorchScript SDK backend

### LINKED ISSUE #487: [Feature] wrap libtorch as net module in MMDeploy SDK
**Describe the feature**
As described in the title, this feature requests developing NetModule by wrapping libtorch in MMDeploy SDK.
Therefore, we can do torchscript model  inference via SDK API

**Motivation**
#345

**Related resources**

**Additional context**
If you would like to implement the feature and create a PR, please leave a comment here and that would be much appreciated.

lzhangzz:
Resolves #487 

- supports torch 1.8~1.12

Known issues:
1. torchvision is required and must be manually built for some operators to work (e.g. NMS)
2. `LD_LIBRARY_PATH` to libtorch libraries must be set in monolithic build
3. `-D_GLIBCXX_USE_CXX11_ABI=0` is required for all C++ dependencies on Ubuntu 16 and later


csrc/mmdeploy/net/torchscript/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+
+project(mmdeploy_torch_net)
+
+find_package(Torch REQUIRED)
+find_package(TorchVision)
```
lvhan028, line 6: In what case `torchvision` is needed?

csrc/mmdeploy/net/torchscript/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+target_link_libraries(${PROJECT_NAME} PRIVATE
+        ${TORCH_LIBRARIES})
+
+target_link_directories(${PROJECT_NAME} INTERFACE
+        $<BUILD_INTERFACE:${Torch_DIR}/../../../lib>)
+
+target_link_libraries(${PROJECT_NAME} PRIVATE
+        mmdeploy_torchscript_ops_obj)
+
+if ("cuda" IN_LIST MMDEPLOY_TARGET_DEVICES)
```
grimoire, line 19: `MMDEPLOY_USE_CUDA` could be placed outside `torchscript/CMakeLists.txt` since it might be used by other backend or transforms.

csrc/mmdeploy/net/torchscript/torch_net.cpp
```diff
@@ -0,0 +1,236 @@
...
+        OUTCOME_TRY(output_tensor_[index],
+                    FromTorchTensor(x.toTensor(), output_tensor_[index].name()));
+        ++index;
+      }
+    } else {
+      MMDEPLOY_ERROR("{}", toString(outputs.type()));
+      return Status(eNotSupported);
+    }
+  } catch (const std::exception& e) {
+    MMDEPLOY_ERROR("unhandled exception: {}", e.what());
```
lvhan028, line 207: May return failure in exceptional case

csrc/mmdeploy/net/torchscript/CMakeLists.txt
lzhangzz: when model contains operators from torchvision, like `torchvision.ops.nms`.

lvhan028:
I got build error
```
../../lib/libmmdeploy_torch_net.so: undefined reference to `spdlog::logger::err_handler_(std::string const&)'
../../lib/libmmdeploy_torch_net.so: undefined reference to `mmdeploy::Model::GetModelConfig(std::string const&) const'
../../lib/libmmdeploy_torch_net.so: undefined reference to `mmdeploy::Model::ReadFile(std::string const&)'
../../lib/libmmdeploy_torch_net.so: undefined reference to `fmt::v8::vformat(fmt::v8::basic_string_view<char>, fmt::v8::basic_format_args<fmt::v8::basic_format_context<fmt::v8::appender, char> >)'
../../lib/libmmdeploy_torch_net.so: undefined reference to `mmdeploy::format_value(mmdeploy::Value const&)'
collect2: error: ld returned 1 exit status
```

lvhan028 (approved):

grimoire (approved):
LGTM

PR #899, opened 2022-08-14
[Enhancement] Rewriter support pre-import function

grimoire:
test script:

```python
import torch
from torch import matmul

try:
    # disable isort
    from mmdeploy.core import FUNCTION_REWRITER, RewriterContext
except Exception:
    pass


@FUNCTION_REWRITER.register_rewriter('torch.matmul')
def matmul__custom(ctx, input, other):
    print('Rewriter enabled')
    return ctx.origin_func(input, other)


print('xxx before')
with RewriterContext():
    x = torch.rand(4, 4)
    y = torch.rand(4, 4)

    matmul(x, y)

print('xxx after')
matmul(x, y)

```


RunningLeon (approved):
LGTM

hanrui1sensetime (approved):
LGTM.

PR #907, opened 2022-08-17
Sync v0.7.0 to dev-1.x

RunningLeon:
## Motivation

Merge v0.7.0 to dev-1.x branch

## Modification

Fix conflicts.

## BC-breaking (Optional)


grimoire (approved):
LGTM

PR #912, opened 2022-08-17
[Fix] fix device error in dump-info

### LINKED ISSUE #909: Accrording the get_started.md, it occur the problem: cpu is invalid for the backend tensort
Hello, this is my first time to use the mmdeploy, i follow the [get_started.md](https://github.com/open-mmlab/mmdeploy/blob/master/docs/en/get_started.md) to convert model with _Linux-x86_64, CUDA 11.x, TensorRT 8.2.5.1_, just like the **md** show 
image

but it occur the problem: **ValueError: cpu is invalid for the backend tensort**.

By the way , my env as follow:
40A687CFDC1C729D708693F8848FAC0D
0F13DD7514ED2B443CBBD92E2C0D5D3B

AllentDan:
Fix #909


grimoire (approved):

RunningLeon (approved):
LGTM

PR #947, opened 2022-08-25
fix(backend): disable cublaslt for cu102

tpoisonooo:

## Motivation

fix model convert crash for cu102+trt8+mmdet3d centerpoint

## Modification

* Update doc to install cu102 patch
* Disable cublaslt heuristic for cu102

## Checklist

### DIFF
--- a/mmdeploy/backend/tensorrt/utils.py
+++ b/mmdeploy/backend/tensorrt/utils.py
@@ -94,7 +94,7 @@ def from_onnx(onnx_model: Union[str, onnx.ModelProto],
 
     load_tensorrt_plugin()
     # create builder and network
-    logger = trt.Logger(trt.Logger.INFO)
+    logger = trt.Logger(log_level)
     builder = trt.Builder(logger)
     EXPLICIT_BATCH = 1 << (int)(
         trt.NetworkDefinitionCreationFlag.EXPLICIT_BATCH)
--- a/mmdeploy/backend/tensorrt/utils.py
+++ b/mmdeploy/backend/tensorrt/utils.py
@@ -42,7 +42,7 @@ def load(path: str) -> trt.ICudaEngine:
 
 
 def search_cuda_version() -> str:
-    """try import torch.version.cuda, then try use cmd.
+    """try cmd to get cuda version, then try `torch.cuda`
 
     Returns:
         str: cuda version, for example 10.2
--- a/mmdeploy/backend/tensorrt/utils.py
+++ b/mmdeploy/backend/tensorrt/utils.py
@@ -42,7 +42,7 @@ def search_cuda_version() -> str:
     """try import torch.version.cuda, then try use cmd.
 
     Returns:
-        str: _description_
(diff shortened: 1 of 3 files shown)


mmdeploy/backend/tensorrt/utils.py
```diff
@@ -118,6 +118,17 @@ def from_onnx(onnx_model: Union[str, onnx.ModelProto],
 
     config = builder.create_builder_config()
     config.max_workspace_size = max_workspace_size
+    try:
+        import torch
```
grimoire, line 122: I am not sure if it is a good idea to get CUDA version from torch.
This API is expected to be able to use without PyTorch. Do we have other ways to check the CUDA version?

tpoisonooo: I have no better idea

tpoisonooo: Some one use mmdeploy  -->  guy has installed mmxxx repo ---> pytorch already installed.

grimoire: Users might export ONNX on host A and convert to TensorRT on host B (without PyTorch). And the PyTorch has it's builtin CUDA which might be different from the one in the system.
Can we use `ctypes` to load runtime api for the version check?

```python
from ctypes import cdll, pointer, c_int32
dll = cdll.LoadLibrary('libcudart.so')
version = c_int32(0)
dll.cudaRuntimeGetVersion(pointer(version))
print(version)
```

lvhan028: > I have no better idea. TRT input comes from torch, and torch depends on cuda... I think import torch is a must.

I didn't get `TRT input comes from torch`

RunningLeon (approved):
LGTM

mmdeploy/backend/tensorrt/utils.py
tpoisonooo: Have you ever deploy model with mmdeploy without  torch and `mmcls` ?

lvhan028 (approved):

lvhan028:
ci error in "backend-ort / script_install (3.7) (pull_request)"

mmdeploy/backend/tensorrt/utils.py
```diff
@@ -38,6 +38,55 @@ def load(path: str) -> trt.ICudaEngine:
...
+def search_cuda_version() -> str:
+    """try import torch.version.cuda, then try use cmd.
+
+    Returns:
+        str: cuda version, for example 10.2
+    """
+
+    version = None
+    try:
+        import torch
```
grimoire, line 50: Does not prefer this way.
Here is the check env in PyTorch. We can share the same method.
https://github.com/pytorch/pytorch/blob/3ac9f6683dc8f17e030699da4df6c767f22939b6/torch/utils/collect_env.py#L347

grimoire (approved):
LGTM

PR #958, opened 2022-08-30
fix regression test 

RunningLeon:
## Motivation

Please describe the motivation of this PR and the goal you want to achieve through this PR.

## Modification

fix regression test failed when 
1. metafile not exists
2. running `test.py` with sdk + ort(cuda device).
3. running `test.py` with sdk + trt(device=cuda:0)

## BC-breaking (Optional)
None

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdeploy/backend/sdk/wrapper.py
+++ b/mmdeploy/backend/sdk/wrapper.py
@@ -1,7 +1,7 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import mmdeploy_python as c_api
 
-from mmdeploy.utils import Backend, parse_device_id
+from mmdeploy.utils import Backend, parse_device_id, parse_device_type
 from mmdeploy.utils.timer import TimeCounter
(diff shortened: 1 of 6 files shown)


AllentDan (approved):
LGTM

mmdeploy/backend/sdk/wrapper.py
```diff
@@ -12,8 +12,11 @@ class SDKWrapper(BaseWrapper):
     def __init__(self, model_file, task_name, device):
         super().__init__([])
         creator = getattr(c_api, task_name)
-        # TODO: get device id somewhere
-        self.handle = creator(model_file, device, 0)
+        device_id = parse_device_id(device)
+        device_type = 'cpu' if device_id == -1 else 'cuda'
```
lzhangzz, line 16: Device specific names should not appear in device agnostic code. There will be other device name some (e.g. "npu" from Ascend devices). Prefer splitting device name & index and pass device name as it is.

lzhangzz (approved):
LGTM

PR #1002, opened 2022-09-05
[Fix] Add isolated option for TorchScript SDK backend

lzhangzz:
Add isolated option for TorchScript SDK backend since it's causing ABI issues hard to resolve.

### DIFF
--- a/csrc/mmdeploy/net/torchscript/CMakeLists.txt
+++ b/csrc/mmdeploy/net/torchscript/CMakeLists.txt
@@ -2,23 +2,27 @@
 
 project(mmdeploy_torch_net)
 
-find_package(Torch REQUIRED)
-find_package(TorchVision)
+option(MMDEPLOY_TORCHSCRIPT_SDK_BACKEND "Build TorchScript SDK backend" OFF)
 
-mmdeploy_add_module(${PROJECT_NAME} torch_net.cpp)
+if (MMDEPLOY_TORCHSCRIPT_SDK_BACKEND)
+    find_package(Torch REQUIRED)
+    find_package(TorchVision QUIET)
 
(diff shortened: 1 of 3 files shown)


csrc/mmdeploy/net/torchscript/CMakeLists.txt
```diff
@@ -2,23 +2,27 @@
 
 project(mmdeploy_torch_net)
 
-find_package(Torch REQUIRED)
-find_package(TorchVision)
+option(MMDEPLOY_TORCHSCRIPT_SDK_BACKEND "Build TorchScript SDK backend" OFF)
```
RunningLeon, line 5: may update doc as well.

lvhan028, line 5: Can `MMDEPLOY_BUILD_SDK=ON` and `MMDEPLOY_TARGET_BACKENDS="torchscript"` be enough?

lzhangzz: If you just want the converter to work with torchscript (for optimization passes), this would introduce the chaos of different C++ ABIs.

lvhan028: If only converter is concerned, we can disable `MMDEPLOY_BUILD_SDK`

lzhangzz: What if I need to use converter with optimizer to convert TensorRT models and use SDK with TensorRT backend?

lvhan028 (approved):

RunningLeon (approved):
LGTM

PR #1018, opened 2022-09-07
fix mmdeploy builder on windows

irexyc:
The last PR forget to use `osp.join` and `_remove_if_exist` should call after `_call_command`

Fix https://github.com/open-mmlab/mmdeploy/pull/960

### DIFF
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -1,4 +1,3 @@
-pyyaml
 asynctest
 coverage
 flake8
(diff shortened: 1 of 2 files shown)


RunningLeon (commented):
Could you add `pyyaml` in `requirements/test.txt` ?

RunningLeon (approved):
LGTM

grimoire (approved):

Issue #1028, opened 2022-09-09
Core ML: Could not export Python function call 'RoIAlignFunction'


Typiqally:
When trying to export `faster_rcnn_regnetx-3.2GF_fpn_1x_coco` I get the following:

## Command
```
python mmdeploy/tools/deploy.py \
    mmdeploy/configs/mmdet/detection/detection_coreml_static-800x1344.py \
    mmdetection/configs/regnet/faster_rcnn_regnetx-3.2GF_fpn_1x_coco.py \
    checkpoints/faster_rcnn_regnetx-3.2GF_fpn_1x_coco_20200517_175927-126fd9bf.pth \
    mmdetection/demo/demo.jpg \
    --work-dir work_dir/faster_rcnn_regnetx \
    --device cpu
```

## Result
```
2022-09-09 12:32:24,161 - mmdeploy - INFO - Start pipeline mmdeploy.apis.pytorch2torchscript.torch2torchscript in subprocess
load checkpoint from local path: checkpoints/faster_rcnn_regnetx-3.2GF_fpn_1x_coco_20200517_175927-126fd9bf.pth
/Users/typically/Workspace/vbti-plant-morphology/mmdetection/mmdet/datasets/utils.py:66: UserWarning: "ImageToTensor" pipeline is replaced by "DefaultFormatBundle" for batch inference. It is recommended to manually replace it in the test data pipeline in your config file.
  warnings.warn(
2022-09-09 12:32:25,076 - mmdeploy - INFO - Export PyTorch model to torchscript.
/Users/typically/Workspace/vbti-plant-morphology/mmdeploy/mmdeploy/core/rewriters/rewriter_utils.py:255: UserWarning: Detect multiple valid rewriters formmdeploy.codebase.mmdet.core.post_processing._multiclass_nms, use the first rewriter.
  warnings.warn(
/opt/homebrew/Caskroom/miniconda/base/envs/mmlabs/lib/python3.8/site-packages/torch/nn/functional.py:718: UserWarning: Named tensors and all their associated APIs are an experimental feature and subject to change. Please do not use them for anything important until they are released as stable. (Triggered internally at  /tmp/pip-req-build-jhk0fkj8/c10/core/TensorImpl.h:1156.)
  return torch.max_pool2d(input, kernel_size, stride, padding, dilation, ceil_mode)
[...]

PR #1029, opened 2022-09-09
Fix `RoIAlignFunction` error for CoreML backend

Typiqally:
## Motivation

The CoreML backend would produce an error when exporting a model that uses the `RoIAlignFunction`, see issue #1028. This was already fixed in #724, but only for the TorchScript backend specifically.

## Modification

Since the CoreML backend uses the TorchScript backend as a first step, the condition in https://github.com/open-mmlab/mmdeploy/blob/master/mmdeploy/codebase/mmdet/models/roi_heads/single_level_roi_extractor.py has to be updated to support the CoreML backend.


grimoire (approved):
Thanks for the contribution! LGTM

irexyc (approved):

PR #1006, opened 2022-09-05
SDK ocr 2.0

AllentDan:
There are bc breakings at DBNet and CRNN for SDK. The post-process logic was changed in OCR 2.0.

PR #1040, opened 2022-09-13
[2.0] Refactor dump info

AllentDan:
Based on PR `SDK ocr 2.0` #1006. 
Mainly changed:
1. move get_preprocess into task_processor
2. move get_postprocess into task_processor
3. refactor func `get_models`

### DIFF
--- a/mmdeploy/codebase/base/task.py
+++ b/mmdeploy/codebase/base/task.py
@@ -322,7 +322,7 @@ class BaseTask(metaclass=ABCMeta):
         return input_data['inputs']
 
     @abstractmethod
-    def get_preprocess(self) -> Dict:
+    def get_preprocess(self, *args, **kwargs) -> Dict:
         """Get the preprocess information for SDK.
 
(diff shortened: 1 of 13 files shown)


RunningLeon:
@AllentDan may need to fix conflicts

mmdeploy/codebase/base/task.py
```diff
@@ -334,7 +334,7 @@ def get_postprocess(self) -> Dict:
         pass
 
     @abstractmethod
-    def get_model_name(self) -> str:
+    def get_model_name(self, *args, **kwargs) -> str:
```
RunningLeon, line 337: it seems no need to add this two arguments if all you need is in self.model_cfg.

AllentDan: For OCR, `work_dir` is used. And we'd better add this since we don't know if we will add more args in the future.

RunningLeon (approved):
LGTM

PR #1042, opened 2022-09-13
Update check_env.py

shuxp:
collect_env moved mmcv to mmengine in branch 1.x
just correct library import.


shuxp:
bug fix in issue#979 , which is in reviewing. close this pr.

PR #1045, opened 2022-09-14
Rewrite Conv2dAdaptiveOps for conversion of EfficientNet (static shape)

SsTtOoNnEe:
## Motivation

Rewrite Conv2dAdaptiveOps for conversion of EfficientNet (static shape). In the PyTorch version of 1.10 (or below), the conversion of EfficientNet can not be fully completed because of different ONNX parsing behaviours against the higher PyTorch versions (>=1.10).  In detail, the operator of `Conv2dAdaptivePadding` in MMCV (which is used in EfficientNet) conducts padding shape computation before the convolution, which introduces complex and redundant nodes in the ONNX graph. By using NVIDIA Polygraphy tools, the output ONNX graph can be simplified to allow for the following conversion procedure. Alternatively, in order to maintain the robustness of MMDeploy, the `Conv2dAdaptivePadding` operator can be rewritten given different configurations (eg, dynamic and static shapes).

## Modification
For the static model conversion task, we extracted the padding computation from the `Cov2dAdaptivePadding` operator to encapsulate it into a separate `torch.autograd.Fcuntion` class. After the modification, the ONNX graph exported from the adaptive convolution will not include the computation of padding, thus making it static.

## Test results
| PyTorch version | Top1-Accuracy | Top5-Accuracy |
| ---- | ---- | ---- |
| 1.12.2 | 75.64 | 92.38 |
| 1.8.0 (`Rewritten`) | 75.12 | 92.24 |


grimoire:
Thanks for the contribution! Could you provide the deploy config you used to convert efficientnet?

mmdeploy/mmcv/cnn/conv2d_adaptive_padding.py
```diff
@@ -0,0 +1,84 @@
...
+@FUNCTION_REWRITER.register_rewriter(
+    func_name='mmcv.cnn.bricks.conv2d_adaptive_padding.Conv2dAdaptivePadding.forward',
+    backend=Backend.TENSORRT.value)
+def conv2d_adaptive_padding__forward__tensorrt(ctx, self, x):
+    """Implementation of 2D Convolution in tensorflow with `padding` as "same",
+    which applies padding to input (if needed) so that the input image gets fully 
+    covered by filter and stride you specify. For stride 1, this will ensure that the
+    output image size is the same as input. For stride 2, output image size will be half.
+    
+    Args:
```
RunningLeon, line 69: could only add argument of `ctx`, `self` and `x`

mmdeploy/mmcv/cnn/conv2d_adaptive_padding.py
```diff
@@ -0,0 +1,84 @@
...
+            mode_s="constant",
+            outputs=1
+        )
+
+
+@FUNCTION_REWRITER.register_rewriter(
+    func_name='mmcv.cnn.bricks.conv2d_adaptive_padding.Conv2dAdaptivePadding.forward',
+    backend=Backend.TENSORRT.value)
+def conv2d_adaptive_padding__forward__tensorrt(ctx, self, x):
+    """Implementation of 2D Convolution in tensorflow with `padding` as "same",
```
RunningLeon, line 64: may add reason and what changes here

mmdeploy/mmcv/cnn/conv2d_adaptive_padding.py
```diff
@@ -0,0 +1,84 @@
...
+        padding (int or tuple, optional): Zero padding added to both sizes of 
+            the input. Default: 0
+        dilation (int or tuple, optional): spacing between kernel elements.
+            Default: 1
+        groups (int, optional): Number of blocked connections from input channels to
+            output channels. Default: 1
+        bias (bool, optional): If ``True``, add a learnable bias to the output.
+            Default: ``True``
+    """
+    x = AdaptivePadOp.apply(x, self.weight.shape[2:], self.stride, self.dilation)
```
RunningLeon, line 83: here we only support static shape, so we need to check if its static by referring to https://github.com/open-mmlab/mmdeploy/blob/1c5dcb799badbad717f6bfed615acb4a8e52f92f/mmdeploy/codebase/mmseg/models/decode_heads/aspp_head.py#L26

lvhan028:
Please resolve the lint error. You can use `pre-commit` to check if there is lint issues before committing like below, 

```shell
python3 -m pip install pre-commit
cd /the/root/path/of/mmdeploy
pre-commit install
pre-commit run --all-files
```

RunningLeon:
@SsTtOoNnEe may add unit test for this rewriting.

mmdeploy/mmcv/cnn/conv2d_adaptive_padding.py
SsTtOoNnEe: Got it. Thanks

mmdeploy/mmcv/cnn/conv2d_adaptive_padding.py
SsTtOoNnEe: Sure. Thanks

mmdeploy/mmcv/cnn/conv2d_adaptive_padding.py
SsTtOoNnEe: Done

SsTtOoNnEe:
@RunningLeon I have modified the PR according to your comments. Please kindly review it.

mmdeploy/mmcv/cnn/conv2d_adaptive_padding.py
```diff
@@ -0,0 +1,99 @@
...
+        Tensor: forward result of 2D convolution after padding
+    """
+
+    deploy_cfg = ctx.cfg
+    is_dynamic_flag = is_dynamic_shape(deploy_cfg)
+    if not is_dynamic_flag:
+        x = AdaptivePadOp.apply(x, self.weight.shape[2:], self.stride,
+                                self.dilation)
+    else:
+        img_h, img_w = x.size()[-2:]
```
RunningLeon, line 83: could use `ctx.origin_func` that is the original function.
```
else:
    x = ctx.origin_func(x)
return x

Comment (SsTtOoNnEe) [reply mmdeploy/mmcv/cnn/conv2d_adaptive_padding.py:83]: Noted. Thanks.

File: mmdeploy/mmcv/cnn/__init__.py [hunk 5]
```diff
@@ -1,5 +1,10 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+from .conv2d_adaptive_padding import (
```
RunningLeon, line 2: @SsTtOoNnEe  hi, could update the supported model config in the regression [yaml](https://github.com/open-mmlab/mmdeploy/blob/master/tests/regression/mmcls.yml)?

tests/regression/mmcls.yml
```diff
@@ -51,7 +51,7 @@ tensorrt:
   pipeline_trt_static_fp16: &pipeline_trt_static_fp16
     convert_image: *convert_image
     backend_test: *default_backend_test
-    deploy_config: configs/mmcls/classification_tensorrt-fp16_static-224x224.py
+    deploy_config: configs/mmcls/ll -fp16_static-224x224.py
```
RunningLeon, line 54: something wrong here?

SsTtOoNnEe: Corrected.

RunningLeon:
@SsTtOoNnEe Hi, could you fix the conflicts?

RunningLeon:
@SsTtOoNnEe Pls. fix lint error.

RunningLeon (approved):
LGTM

RunningLeon:
@grimoire any comment on this PR?

grimoire (approved):
LGTM

PR #1053, opened 2022-09-15
fix ort-gpu inference

irexyc:
## Motivation

When use onnxruntime-gpu, it seems we need to synchronize stream before run session.

### DIFF
--- a/mmdeploy/backend/onnxruntime/wrapper.py
+++ b/mmdeploy/backend/onnxruntime/wrapper.py
@@ -92,6 +92,8 @@ class ORTWrapper(BaseWrapper):
         for name in self._output_names:
             self.io_binding.bind_output(name)
         # run session to get outputs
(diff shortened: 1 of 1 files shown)


grimoire:
Is it possible to set the cuda stream in ONNX Runtime `torch.cuda.current_stream().cuda_stream` ?

irexyc:
> Is it possible to set the cuda stream in ONNX Runtime `torch.cuda.current_stream().cuda_stream` ?

It can pass stream to provider options in c api, but there seems no python api to do that.
https://github.com/microsoft/onnxruntime/blob/v1.12.0/onnxruntime/core/providers/cuda/cuda_execution_provider_info.cc#L43-L103

grimoire (approved):

tpoisonooo (approved):
LGTM.

PR #1094, opened 2022-09-23
Add mmseg configs for compatibility with Fast-SCNN model in ncnn backend

lansfair:

## Motivation
Add Fast-SCNN mmsgemention quantization evaluation result for ncnn

## Modification
Implementation details：https://zhuanlan.zhihu.com/p/567319765
1. fix interpolate__ncnn bug
This may have rounding errors before modification
2. fix adaptive_avg_pool2d_ncnn bug
For ncnn, AdaptiveAvgPool2d doesn't seem to be equal to pytorch's nn.AdaptiveAvgPool2d when the output size is 1
3. Add Fast-SCNN mmsgemention quantization evaluation result for ncnn

## BC-breaking (Optional)

No

## Use cases (Optional)

## Checklist

3. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
7. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
8. The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/docs/en/03-benchmark/quantization.md
+++ b/docs/en/03-benchmark/quantization.md
@@ -41,3 +41,11 @@ Note: MMPose models are tested with `flip_test` explicitly set to `False` in mod
 | :-----------------------------------------------------------------------------------------------------------------: | :-----: | :------------: | :------------: |
 | [EDSRx2](https://github.com/open-mmlab/mmediting/blob/master/configs/restorers/edsr/edsr_x2c64b16_g1_300k_div2k.py) |  Set5   | 35.7733/0.9365 | 35.4266/0.9334 |
 | [EDSRx4](https://github.com/open-mmlab/mmediting/blob/master/configs/restorers/edsr/edsr_x4c64b16_g1_300k_div2k.py) |  Set5   | 30.2194/0.8498 | 29.9340/0.8409 |
+
+### mmseg
+
+|                                                             model                                                              |  dataset   | fp32 mIoU | int8 mIoU |
+| :----------------------------------------------------------------------------------------------------------------------------: | :--------: | :-------: | :-------: |
+| [Fast-SCNN](https://github.com/open-mmlab/mmsegmentation/blob/master/configs/fastscnn/fast_scnn_lr0.12_8x4_160k_cityscapes.py) | cityscapes |   70.96   |   70.24   |
+
+Note:Int8 models of the Fast-SCNN requires ncnnoptimize.                                    |            |           |           |
--- a/docs/en/03-benchmark/quantization.md
+++ b/docs/en/03-benchmark/quantization.md
@@ -48,4 +48,7 @@ Note: MMPose models are tested with `flip_test` explicitly set to `False` in mod
 | :----------------------------------------------------------------------------------------------------------------------------: | :--------: | :-------: | :-------: |
 | [Fast-SCNN](https://github.com/open-mmlab/mmsegmentation/blob/master/configs/fastscnn/fast_scnn_lr0.12_8x4_160k_cityscapes.py) | cityscapes |   70.96   |   70.24   |
 
-Note:Int8 models of the Fast-SCNN requires ncnnoptimize.                                    |            |           |           |
+Note:
+
+- Int8 models of the Fast-SCNN requires ncnnoptimize.
+- NCNN will extract 512 images from the train as a calibration dataset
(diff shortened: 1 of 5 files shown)


tpoisonooo (approved):
LGTM.

tpoisonooo:
How to fix lint:
```bash
$ python3 -m pip install pre-commit
$ pre-commit install .
$ pre-commit run --all-files
```

hanrui1sensetime (commented):
Please fix unittest which is failed but skipped by CI:  `tests/test_pytorch`

RunningLeon:
@lansfair Hi, could you resolve conflicts?

lansfair:
> @lansfair Hi, could you resolve conflicts?
I already resolve it.

mmdeploy/pytorch/functions/interpolate.py
```diff
@@ -27,9 +27,8 @@ def interpolate__ncnn(ctx,
 
     input_size = input.shape
     if scale_factor is None:
-        scale_factor = [
-            s_out / s_in for s_out, s_in in zip(size, input_size[2:])
-        ]
+        scale_factor = [(s_out / s_in + 0.00001)
```
RunningLeon, line 30: why do we add this small value, this rewriting would affect all models.

lansfair: Sorry，the last commit committed the wrong version，just add this small value the scale_factor are not integers. Therefore, there may be a rounding error. For example, if the size is 6x6, the resize is 32x32, and the scale_factor is 5.333333333... , the number of bits that can be retained during the conversion is limited, so the size of resize becomes 31.999999 when computed with scale_factor, which is rounded down to 31x31.

docs/en/03-benchmark/quantization.md
```diff
@@ -41,3 +41,11 @@ Note: MMPose models are tested with `flip_test` explicitly set to `False` in mod
...
 | [EDSRx2](https://github.com/open-mmlab/mmediting/blob/master/configs/restorers/edsr/edsr_x2c64b16_g1_300k_div2k.py) |  Set5   | 35.7733/0.9365 | 35.4266/0.9334 |
 | [EDSRx4](https://github.com/open-mmlab/mmediting/blob/master/configs/restorers/edsr/edsr_x4c64b16_g1_300k_div2k.py) |  Set5   | 30.2194/0.8498 | 29.9340/0.8409 |
+
+### mmseg
+
+|                                                             model                                                              |  dataset   | fp32 mIoU | int8 mIoU |
+| :----------------------------------------------------------------------------------------------------------------------------: | :--------: | :-------: | :-------: |
+| [Fast-SCNN](https://github.com/open-mmlab/mmsegmentation/blob/master/configs/fastscnn/fast_scnn_lr0.12_8x4_160k_cityscapes.py) | cityscapes |   70.96   |   70.24   |
+
+Note:Int8 models of the Fast-SCNN requires ncnnoptimize.                                    |            |           |           |
```
RunningLeon, line 51: Hi, could you add note here how many images were used in calibration dataset?

mmdeploy/pytorch/functions/interpolate.py
RunningLeon: @lansfair Hi, could we rewrite `resize` and `Upsample` for in mmseg [code](https://github.com/open-mmlab/mmsegmentation/blob/master/mmseg/ops/wrappers.py) to use `size` instead of `scale factor` for ncnn. It seems resize input is `size` for instance in [here](https://github.com/open-mmlab/mmsegmentation/blob/a68e7c2823f6c9c79eb2fea68ecbf626aebd2b85/mmseg/models/segmentors/encoder_decoder.py#L76)

mmdeploy/pytorch/functions/interpolate.py
```diff
@@ -28,7 +28,9 @@ def interpolate__ncnn(ctx,
     input_size = input.shape
     if scale_factor is None:
         scale_factor = [
-            s_out / s_in for s_out, s_in in zip(size, input_size[2:])
+            s_out / s_in if int(s_out / s_in) == s_out / s_in else
+            (s_out / s_in + 0.00001)
```
RunningLeon, line 32: @lansfair  Hi, pls. add a comment here for the change

RunningLeon (approved):
LGTM after resolving the comments

PR #938, opened 2022-08-23
[Enhancement] Refactor SDK pipeline 

lzhangzz:
usage

`det_cls.cpp`
`det_pose.cpp`
`text_det_recog.cpp`

PR #1145, opened 2022-09-30
[Feature] Add RKNN support to SDK

AllentDan:


lvhan028:
May complete the following tasks:
- add github action ci to build mmdeploy using rknn-toolchain
- add `en/01-how-to-build/rknn.md` and `zh_cn/01-how-to-build/rknn.md`  to introduce how to build and install mmdeploy on RK
-

csrc/mmdeploy/net/rknn/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+
+project(mmdeploy_rknn_net)
+
+add_library(rknn SHARED IMPORTED)
+
+if (NOT EXISTS $ENV{RKNN_ROOT}/runtime/RK3588/Linux/librknn_api/aarch64/)
```
lvhan028, line 7: May put them in a toolchain cmake file like `cmake/toolchains/riscv64-unknown-linux-gun.cmake` did.

csrc/mmdeploy/net/rknn/rknn_net.h
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) OpenMMLab. All rights reserved.
+
+#ifndef MMDEPLOY_SRC_NET_RKNN_RKNN_NET_H_
+#define MMDEPLOY_SRC_NET_RKNN_RKNN_NET_H_
+
+#include "mmdeploy/core/mpl/span.h"
+#include "mmdeploy/core/net.h"
+#include "rknn_api.h"
+
+namespace mmdeploy {
```
lvhan028, line 10: namespace mmdeploy -> namespace mmdeploy::framework.

namespace mmdeploy is used for C++ API

For inner C++ classes, we use `namespace mmdeploy::framework`. Please refer to #938

csrc/mmdeploy/core/CMakeLists.txt
```diff
@@ -65,7 +65,7 @@ if (MMDEPLOY_BUILD_SDK_CXX_API)
 endif ()
 
 target_link_libraries(${PROJECT_NAME} PUBLIC ${SPDLOG_LIB})
-if (NOT (MSVC OR ANDROID OR APPLE))
+if (NOT (MSVC OR ANDROID OR APPLE OR (CMAKE_C_COMPILER MATCHES "aarch64")))
```
lvhan028, line 68: I think the method from 'pybind11/tests/CMakeLists.txt' to set `STD_FS_LIB` is better than ours. @lzhangzz what's your opinion?

AllentDan:
Since rknpu2 itself does not maintain the link for the cross compiler and I used this [link](https://gitlab.com/firefly-linux/prebuilts/gcc/linux-x86/aarch64/gcc-buildroot-9.3.0-2020.03-x86_64_aarch64-rockchip-linux-gnu/-/tree/firefly/aarch64-rockchip-linux-gnu). Should I copy it to my own account like [this](https://github.com/open-mmlab/mmdeploy/blob/b2cb5722f732c4789a11e282d13b00b6e25e2b0d/.github/workflows/linux-riscv64-gcc.yml#L40)? @lvhan028

csrc/mmdeploy/net/rknn/CMakeLists.txt
AllentDan: Done

csrc/mmdeploy/net/rknn/rknn_net.h
AllentDan: Done

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  // Get Model Input Output Info
+  rknn_input_output_num io_num;
+  ret = rknn_query(ctx_, RKNN_QUERY_IN_OUT_NUM, &io_num, sizeof(io_num));
+  if (ret != RKNN_SUCC) {
+    MMDEPLOY_ERROR("rknn_query fail! ret= {}", ret);
+    return Status(eFail);
+  }
+  printf("model input num: %d, output num: %d\n", io_num.n_input, io_num.n_output);
```
lvhan028, line 77: Please use `MMDEPLOY_INFO` or `MMDEPLOY_DEBUG` to print logs.

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,211 @@
...
+  printf(
+      "  index=%d, name=%s, n_dims=%d, dims=[%d, %d, %d, %d], n_elems=%d, size=%d, fmt=%s, "
+      "type=%s, qnt_type=%s, "
+      "zp=%d, scale=%f\n",
+      attr->index, attr->name, attr->n_dims, attr->dims[0], attr->dims[1], attr->dims[2],
+      attr->dims[3], attr->n_elems, attr->size, get_format_string(attr->fmt),
+      get_type_string(attr->type), get_qnt_type_string(attr->qnt_type), attr->zp, attr->scale);
+}
+
+unsigned char* RKNNNet::load_model(const char* filename, int* model_size) {
```
lvhan028, line 27: This function seems not in use, does it?
If so, may remove it.

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,211 @@
...
+  rknn_tensor_attr input_attrs[io_num.n_input];
+  memset(input_attrs, 0, sizeof(input_attrs));
+  for (int i = 0; i < io_num.n_input; i++) {
+    input_attrs[i].index = i;
+    ret = rknn_query(ctx_, RKNN_QUERY_INPUT_ATTR, &(input_attrs[i]), sizeof(rknn_tensor_attr));
+    if (ret != RKNN_SUCC) {
+      MMDEPLOY_ERROR("rknn_query fail! ret= {}", ret);
+      return Status(eFail);
+    }
+    dump_tensor_attr(&(input_attrs[i]));
```
lvhan028, line 89: May remove the debug code

AllentDan:
> Since rknpu2 itself does not maintain the link for the cross compiler and I used this [link](https://gitlab.com/firefly-linux/prebuilts/gcc/linux-x86/aarch64/gcc-buildroot-9.3.0-2020.03-x86_64_aarch64-rockchip-linux-gnu/-/tree/firefly/aarch64-rockchip-linux-gnu). Should I copy it to my own account like [this](https://github.com/open-mmlab/mmdeploy/blob/b2cb5722f732c4789a11e282d13b00b6e25e2b0d/.github/workflows/linux-riscv64-gcc.yml#L40)? @lvhan028

recap

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,211 @@
...
+    if (ret != RKNN_SUCC) {
+      MMDEPLOY_ERROR("rknn_query fail! ret= {}", ret);
+      return Status(eFail);
+    }
+    dump_tensor_attr(&(output_attrs[i]));
+  }
+  output_attrs_ = std::vector<rknn_tensor_attr>(output_attrs, output_attrs + io_num.n_output);
+
+  for (int i = 0; i < io_num.n_input; ++i) {
+    // TODO get real data type instead of hard code
```
lvhan028, line 108: according to the definition of `rknn_tensor_attr`, there is a member `rknn_tensor_type type`
```
rknn_tensor_type type;      /* the data type of tensor. */
```
May add an inline function to map rknn_tensor_type and mmdeploy tensor data type like MapDataType in `trt_net.cpp`

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,211 @@
...
+Result<void> RKNNNet::Forward() {
+  OUTCOME_TRY(stream_.Wait());
+
+  rknn_input inputs[input_tensors_.size()];
+  memset(inputs, 0, input_tensors_.size() * sizeof(rknn_input));
+  for (int i = 0; i < input_tensors_.size(); i++) {
+    inputs[i].index = i;
+    inputs[i].pass_through = 0;
+    inputs[i].type = RKNN_TENSOR_FLOAT32;
+    inputs[i].fmt = RKNN_TENSOR_NHWC;
```
lvhan028, line 145: What are the transform ops in preprocess? If there are `defaultFormatBundle` or `Image2Tensor`, the data format is `NCHW`

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+  auto name = args["name"].get<std::string>();
+  auto model = context["model"].get<Model>();
+  OUTCOME_TRY(auto config, model.GetModelConfig(name));
+
+  std::string content;
+  OUTCOME_TRY(content, model.ReadFile(config.net));
+  char* model_ptr = const_cast<char*>(content.data());
+  int ret = rknn_init(&ctx_, model_ptr, content.size(), 0, NULL);
+  if (ret < 0) {
```
lvhan028, line 65: better use `if (ret != RKNN_SUCC)`

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,211 @@
...
+    MMDEPLOY_ERROR("Load .rknn failed: {}", config.net);
```
lvhan028, line 66: May print the error code as well.

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,211 @@
...
+    return Status(eInvalidArgument);
+  }
+
+  // Get Model Input Output Info
+  rknn_input_output_num io_num;
+  ret = rknn_query(ctx_, RKNN_QUERY_IN_OUT_NUM, &io_num, sizeof(io_num));
+  if (ret != RKNN_SUCC) {
+    MMDEPLOY_ERROR("rknn_query fail! ret= {}", ret);
```
lvhan028, line 74: May add "RKNN_QUERY_INPUT_ATTR" in the error log as well so that users can know which query failed.

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,211 @@
...
+    return Status(eFail);
+  }
+  printf("model input num: %d, output num: %d\n", io_num.n_input, io_num.n_output);
+
+  printf("input tensors:\n");
+  rknn_tensor_attr input_attrs[io_num.n_input];
+  memset(input_attrs, 0, sizeof(input_attrs));
+  for (int i = 0; i < io_num.n_input; i++) {
+    input_attrs[i].index = i;
+    ret = rknn_query(ctx_, RKNN_QUERY_INPUT_ATTR, &(input_attrs[i]), sizeof(rknn_tensor_attr));
```
lvhan028, line 84: May add "RKNN_QUERY_INPUT_ATTR" in the error log as well so that users can know which query failed.

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,211 @@
...
+  rknn_input_output_num io_num;
+  ret = rknn_query(ctx_, RKNN_QUERY_IN_OUT_NUM, &io_num, sizeof(io_num));
+  if (ret != RKNN_SUCC) {
+    MMDEPLOY_ERROR("rknn_query fail! ret= {}", ret);
+    return Status(eFail);
+  }
+  printf("model input num: %d, output num: %d\n", io_num.n_input, io_num.n_output);
+
+  printf("input tensors:\n");
+  rknn_tensor_attr input_attrs[io_num.n_input];
```
lvhan028, line 80: I don't think defining an array with variable length is a good way. Although GCC extension supports that, not all compiler does it. 
Better to use std::vector directly. We can directly update `input_attrs_`. In this way, we can also avoid `input_attrs_ = std::vector<rknn_tensor_attr>(input_attrs, input_attrs + io_num.n_input);`
```
input_attrs_.resize(io_num.n_input);
for (int i = 0; i < io_num.n_input; ++i) {
   ....
}

```

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,211 @@
...
+    if (ret != RKNN_SUCC) {
+      MMDEPLOY_ERROR("rknn_query fail! ret= {}", ret);
+      return Status(eFail);
+    }
+    dump_tensor_attr(&(input_attrs[i]));
+  }
+  input_attrs_ = std::vector<rknn_tensor_attr>(input_attrs, input_attrs + io_num.n_input);
+
+  printf("output tensors:\n");
+  rknn_tensor_attr output_attrs[io_num.n_output];
```
lvhan028, line 94: same as the comments on `input_attrs`

csrc/mmdeploy/net/rknn/rknn_net.cpp
AllentDan: This is no `defaultFormatBundle` or `Image2Tensor` for rknn now.

csrc/mmdeploy/net/rknn/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+
+project(mmdeploy_rknn_net)
+
+add_library(rknn SHARED IMPORTED)
+
+if( CMAKE_TOOLCHAIN_FILE )
```
lvhan028, line 7: No need to include(CMAKE_TOOLCHAIN_FILE). 
toolchain file is used to specify the cross-compiler and is read early in the CMake run.
The `CMAKE_SYSTEM_PROCESSOR` or `CMAKE_SYSTEM_NAME` can be retrieved directly without including this file

AllentDan: If don't include this, the cmake file could not be found. It is because I gave wrong `CMAKE_SYSTEM_PROCESSOR ` or `CMAKE_SYSTEM_NAME`?

csrc/mmdeploy/net/rknn/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+add_library(rknn SHARED IMPORTED)
+
+if(DEFINED ENV{RKNPU2_DIR})
+    file(TO_CMAKE_PATH $ENV{RKNPU2_DIR} RKNPU2_DIR)
+else()
+    message(FATAL_ERROR "RKNPU2_DIR env must be defined")
+endif()
+
+set_target_properties(rknn PROPERTIES
+  IMPORTED_LOCATION "${RKNPU2_DIR}/runtime/RK3588/Linux/librknn_api/aarch64/librknnrt.so"
```
lvhan028, line 14: I found in `rknpu2` runtime directory, the structure is as following:
```
.
├── Android.bp
├── Android.go
├── init.rknn_server.rc
├── RK356X
├── RK3588
└── RV1106
```
It means that we have to link the corresponding library according to the target device model. But here we hard code the RK3588's runtime path. 
One way is to set the rockchip model in the toolchain file, like `set(CMAKE_SYSTEM_PROCESSOR RK3588)`. 
And we can `set_target_properties` according to `${CMAKE_SYSTEM_PROCESSOR}`

lzhangzz: How about `add_subdirectory(rknn)` unconditionally, and
```cmake
set(SUPPORTED_MODELS RK3588 RK356X RV1126)

foreach (_model_name IN LISTS SUPPORTED_MODELS)
    if (_model_name IN_LIST MMDEPLOY_TARGET_BACKENDS)
        set(MODEL_NAME ${_model_name})
        break ()
    endif ()
endforeach ()

if (MODEL_NAME)
    # add targets
endif ()
```

lvhan028:
> > Since rknpu2 itself does not maintain the link for the cross compiler and I used this [link](https://gitlab.com/firefly-linux/prebuilts/gcc/linux-x86/aarch64/gcc-buildroot-9.3.0-2020.03-x86_64_aarch64-rockchip-linux-gnu/-/tree/firefly/aarch64-rockchip-linux-gnu). Should I copy it to my own account like [this](https://github.com/open-mmlab/mmdeploy/blob/b2cb5722f732c4789a11e282d13b00b6e25e2b0d/.github/workflows/linux-riscv64-gcc.yml#L40)? @lvhan028
> 
> recap

Let's use https://github.com/airockchip/gcc-buildroot-9.3.0-2020.03-x86_64_aarch64-rockchip-linux-gnu. At least this is recommended by rockchip

csrc/mmdeploy/net/rknn/CMakeLists.txt
lvhan028: you mean -DMMDEPLOY_TARGET_BACKENDS="RK3588", right?
Looks better than `set(CMAKE_SYSTEM_PROCESSOR RK3588)`.
I checked https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html, 
`CMAKE_SYSTEM_PROCESSOR`refers to the cpu architecture.

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+  // Set input
+  int ret = rknn_inputs_set(ctx_, input_tensors_.size(), inputs);
+  if (ret < 0) {
+    MMDEPLOY_ERROR("rknn_input_set fail! ret= {}", ret);
+    return Status(eFail);
+  }
+
+  // Get output
+  rknn_output outputs[output_tensors_.size()];
```
lvhan028, line 161: variable length issue of an array

csrc/mmdeploy/net/rknn/rknn_net.cpp
```diff
@@ -0,0 +1,214 @@
...
+  for (size_t i = 0; i < input_shapes.size(); ++i) {
+    input_tensors_[i].Reshape(input_shapes[i]);
+  }
+  return success();
+}
+
+Result<void> RKNNNet::Forward() {
+  OUTCOME_TRY(stream_.Wait());
+
+  rknn_input inputs[input_tensors_.size()];
```
lvhan028, line 142: variable length issue of an array

csrc/mmdeploy/net/rknn/CMakeLists.txt
lvhan028: @lzhangzz if we use RK3588, PR356X as MMDEPLOY_TARGET_BACKENDS arguments, 
exporting SDK model meta info will suffer a lot

AllentDan: Yes, some codes need to be changed. BTW, why can't we add new flags like `MMDEPLOY_TARGET_DEVICE`?

lvhan028: After reconsidering it, `Device` in mmdeploy SDK means it has `buffer`, `stream`, and `event` resources. But npu doesn't have any such concepts. So seems not appropriate to put RK3588 in MMDEPLOY_TARGET_DEVICES. 
In rockchip case, the precise description is device(CPU), backend(rknn), rockchip model(rk3588, ...)
But if we add another build option ROCKCHIP_MODEL, that might be a burden for users.
So I prefer `-DMMDEPLOY_TARGET_BACKENDS="RK3588"`.

csrc/mmdeploy/core/CMakeLists.txt
lzhangzz: #1190

lvhan028 (approved):

lzhangzz (approved):

PR #1164, opened 2022-10-08
[Feature] add support for vipnas

LiuYi-Up:

## Motivation

Add the support for ViPNAS in ncnn(ncnn-int8) backend.

## Modification

Added:  
'mmdeploy/codebase/mmpose/models/heads/vipnas_heatmap_simple_head.py'  
'mmdeploy/mmcv/cnn/context_block.py'  
'mmdeploy/mmcv/cnn/hsigmoid.py'  
'mmdeploy/mmcv/cnn/hswish.py'  

Modified:  
'mmdeploy/pytorch/ops/layer_norm.py'

Docs: some information in 'quantization.md' and 'supported_models.md'

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


mmdeploy/codebase/mmpose/models/heads/vipnas_heatmap_simple_head.py
```diff
@@ -0,0 +1,25 @@
...
+
+    Args:
+    x (list[torch.Tensor[N,K,H,W]]): Input features.
+    flip_pairs (None | list[tuple]):
+        Pairs of keypoints which are mirrored.
+
+    Returns:
+        output_heatmap (torch.Tensor): Output heatmaps.
+    """
+    assert flip_pairs is None
```
tpoisonooo, line 22: why `flip_pairs` must be none ?

tpoisonooo (approved):

mmdeploy/codebase/mmpose/models/heads/vipnas_heatmap_simple_head.py
LiuYi-Up: MMPose models are tested with `flip_test` explicitly set to `False` in model configs. It won't activate the parameter `flip_pairs`.

RunningLeon:
could add unit test for all new rewritings.

mmdeploy/backend/ncnn/wrapper.py
```diff
@@ -102,6 +102,7 @@ def forward(self, inputs: Dict[str,
                 data = input_tensor[batch_id].contiguous()
                 data = data.detach().cpu().numpy()
                 input_mat = ncnn.Mat(data)
+                input_mat = input_mat.clone()
```
RunningLeon, line 105: why do we need to clone here?

tests/test_mmcv/test_mmcv_cnn.py
```diff
@@ -30,3 +32,97 @@ def test_multiheadattention_ncnn():
...
+        deploy_cfg=deploy_cfg,
+        run_with_backend=True)
+    if is_backend_output is None:
+        assert rewrite_outputs is not None
+    else:
+        assert torch.allclose(
+            model_outputs, rewrite_outputs[0], rtol=1e-03, atol=1e-05)
+
+
+if __name__ == '__main__':
```
RunningLeon, line 127: should remove this main function

mmdeploy/backend/ncnn/wrapper.py
LiuYi-Up: When I try to test the `hardsigmoid` and `hardswish` ops which support `forward_inplace` in `ncnn`, something wrong happened with the `refcount` of `input` (it seems that an exception occurred to the memory address result a `segmentation fault`). So I clone this here and it works to solve the problem.

mmdeploy/pytorch/ops/layer_norm.py
```diff
@@ -56,6 +56,25 @@ def _layer_norm_ncnn(g, input, normalized_shape, weight, bias, eps,
     """
     weight.setDebugName('layernorm_weight')
     bias.setDebugName('layernorm_bias')
+    [c, h, w] = normalized_shape
```
RunningLeon, line 59: if `len(normalized_shape)` could be `4`, then this line may break in this case.

RunningLeon (approved):
LGTM after resolving left comments @LiuYi-Up

PR #1190, opened 2022-10-12
[Fix] Detect filesystem library in CMake

lzhangzz:
Also fixed a compile error caused pybind11 when building with clang.


lvhan028 (approved):
LGTM

AllentDan (approved):
LGTM

PR #1238, opened 2022-10-24
[Feature] Support rv1126 in sdk

lvhan028:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily receiving feedbacks. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

support ssd-lite rknn model in MMDeploy SDK

## Modification

After #1203 

## BC-breaking (Optional)

NO

## Use cases (Optional)

If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
3. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #1203, opened 2022-10-14
Add model conversion support to rv1126

AllentDan:

PR #1280, opened 2022-11-01
Add rv1126 yolov3 support to sdk

AllentDan:
After #1203 and #1238


AllentDan:
@lvhan028 @lzhangzz Any comments?

lvhan028:
> @lvhan028 @lzhangzz Any comments?

in test

lvhan028:
I have to do serval modifications when converting yolov3 torch model to rknn-int8 model. 
- Disable `normalize` in yolov3 model's config by changing `std=[255.0, 255.0, 255.0]` to `std=[1.0, 1.0, 1.0]`
- Enable `do_quantization=True` in `rknn.py` and set `mean_value` and `std_value` to `[[0, 0, 0]]` and `[[255.0, 255.0, 255.0]]` respectively
- uncomment the `partition_config` of `yolov3&yolox` in `detection_rknn_static-320x320.py`

I don't think users are willing to buy it.

Let's try to eliminate the manual part.
- Can we move the `normalize` config from torch model config to deploy config according to `do_quantization`?
- How about automatically choosing the proper `partition config` according to `target_platform` and torch model config?

AllentDan:
> I have to do serval modifications when converting yolov3 torch model to rknn-int8 model.
> 
> * Disable `normalize` in yolov3 model's config by changing `std=[255.0, 255.0, 255.0]` to `std=[1.0, 1.0, 1.0]`
> * Enable `do_quantization=True` in `rknn.py` and set `mean_value` and `std_value` to `[[0, 0, 0]]` and `[[255.0, 255.0, 255.0]]` respectively
> * uncomment the `partition_config` of `yolov3&yolox` in `detection_rknn_static-320x320.py`
> 
> I don't think users are willing to buy it.
> 
> Let's try to eliminate the manual part.
> 
> * Can we move the `normalize` config from torch model config to deploy config according to `do_quantization`?
> * How about automatically choosing the proper `partition config` according to `target_platform` and torch model config?

Codes now automatically move the Normalization part for quantization to the rknn model.

Tried to make codes automatically tackle the partition part but failed. The expression is not elegant and it will destroy the rule built before.

csrc/mmdeploy/codebase/mmdet/yolo_head.h
```diff
@@ -0,0 +1,52 @@
...
+  int nms_pre_{1000};
+  float iou_threshold_{0.45f};
+  int min_bbox_size_{0};
+  std::vector<std::vector<std::vector<int>>> anchors_;
+  std::vector<unsigned int> strides_;
+};
+
+class YOLOV3Head : public YOLOHead {
+ public:
+  explicit YOLOV3Head(const Value& cfg);
```
lzhangzz, line 34: `using YOLOHead::YOLOHead;`

csrc/mmdeploy/codebase/mmdet/yolo_head.h
```diff
@@ -0,0 +1,52 @@
...
+  Result<Value> operator()(const Value& prep_res, const Value& infer_res);
+  std::vector<float> yolo_decode(float box_x, float box_y, float box_w, float box_h, int stride,
+                                 std::vector<std::vector<int>> anchor, int j, int i,
+                                 int a) const override;
+};
+
+class YOLOV5Head : public YOLOHead {
+ public:
+  explicit YOLOV5Head(const Value& cfg);
```
lzhangzz, line 43: `using YOLOHead::YOLOHead;`

csrc/mmdeploy/codebase/mmdet/yolo_head.h
```diff
@@ -0,0 +1,52 @@
...
+ public:
+  explicit YOLOHead(const Value& cfg);
+  Result<Value> operator()(const Value& prep_res, const Value& infer_res);
+  int YOLOFeatDecode(const Tensor& feat_map, std::vector<std::vector<int>> anchor, int grid_h,
+                     int grid_w, int height, int width, int stride, std::vector<float>& boxes,
+                     std::vector<float>& objProbs, std::vector<int>& classId,
+                     float threshold) const;
+  Result<Detections> GetBBoxes(const Value& prep_res, const std::vector<Tensor>& pred_maps) const;
+  virtual std::vector<float> yolo_decode(float box_x, float box_y, float box_w, float box_h,
+                                         int stride, std::vector<std::vector<int>> anchor, int j,
```
lzhangzz: `anchor` - pass by const reference to avoid copy

return by `std::array<float, 4>` to void allocation.

csrc/mmdeploy/codebase/mmdet/yolo_head.h
```diff
@@ -0,0 +1,52 @@
...
+#include "mmdeploy/codebase/mmdet/mmdet.h"
+#include "mmdeploy/core/tensor.h"
+
+namespace mmdeploy::mmdet {
+
+class YOLOHead : public MMDetection {
+ public:
+  explicit YOLOHead(const Value& cfg);
+  Result<Value> operator()(const Value& prep_res, const Value& infer_res);
+  int YOLOFeatDecode(const Tensor& feat_map, std::vector<std::vector<int>> anchor, int grid_h,
```
lzhangzz: `anchor` - pass by const reference to avoid copy

csrc/mmdeploy/codebase/mmdet/yolo_head.cpp
```diff
@@ -0,0 +1,229 @@
...
+        }
+      }
+    }
+  }
+  return validCount;
+}
+
+Result<Detections> YOLOHead::GetBBoxes(const Value& prep_res,
+                                       const std::vector<Tensor>& pred_maps) const {
+  std::vector<float> filterBoxes;
```
lvhan028, line 123: inconsistent names. Sometimes in `snake_case` but sometimes in `camelCase`.
Please follow [google c++ style guide](https://google.github.io/styleguide/cppguide.html#Variable_Names) and use `snake_case` for local variables.

csrc/mmdeploy/codebase/mmdet/yolo_head.cpp
```diff
@@ -0,0 +1,229 @@
...
+static float unsigmoid(float y) { return -1.0 * logf((1.0 / y) - 1.0); }
+
+int YOLOHead::YOLOFeatDecode(const Tensor& feat_map, std::vector<std::vector<int>> anchor,
+                             int grid_h, int grid_w, int height, int width, int stride,
+                             std::vector<float>& boxes, std::vector<float>& objProbs,
+                             std::vector<int>& classId, float threshold) const {
+  auto input = const_cast<float*>(feat_map.data<float>());
+  int prop_box_size = feat_map.shape(1) / anchor.size();
+  int kClasses = prop_box_size - 5;
+  int validCount = 0;
```
lzhangzz: snake_case for all variables

configs/mmdet/detection/detection_rknn-fp16_static-320x320.py
```diff
@@ -0,0 +1,34 @@
...
+
+onnx_config = dict(input_shape=[320, 320])
+
+codebase_config = dict(model_type='rknn')
+
+backend_config = dict(
+    input_size_list=[[3, 320, 320]],
+    quantization_config=dict(do_quantization=False))
+
+# # yolov3, yolox for rknn-toolkit and rknn-toolkit2
```
lzhangzz, line 11: remove commented code

configs/mmdet/detection/detection_rknn-int8_static-320x320.py
```diff
@@ -6,24 +6,27 @@
 
 backend_config = dict(input_size_list=[[3, 320, 320]])
 
-# # yolov3, yolox
+# # yolov3, yolox for rknn-toolkit and rknn-toolkit2
```
lzhangzz, line 9: remove commented code

csrc/mmdeploy/codebase/mmdet/yolo_head.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+  std::vector<int> indexArray;
+  for (int i = 0; i < obj_probs.size(); ++i) {
+    indexArray.push_back(i);
+  }
+  Sort(obj_probs, class_id, indexArray);
+
+  Tensor dets(TensorDesc{Device{0, 0}, DataType::kFLOAT,
+                         TensorShape{int(filter_boxes.size() / 4), 4}, "dets"});
+  memcpy(dets.data<float>(), (float*)filter_boxes.data(), filter_boxes.size() * sizeof(float));
```
lzhangzz, line 146: use `std::copy`

csrc/mmdeploy/codebase/mmdet/yolo_head.cpp
```diff
@@ -0,0 +1,229 @@
...
+          box_h = box[3];
+
+          box_x -= (box_w / 2.0);
+          box_y -= (box_h / 2.0);
+          boxes.push_back(box_x);
+          boxes.push_back(box_y);
+          boxes.push_back(box_x + box_w);
+          boxes.push_back(box_y + box_h);
+
+          float maxClassProbs = in_ptr[5 * grid_len];
```
lzhangzz, line 102: use snake_case for all variables

csrc/mmdeploy/codebase/mmdet/yolo_head.h
AllentDan: What's the difference?

csrc/mmdeploy/codebase/mmdet/yolo_head.h
```diff
@@ -0,0 +1,52 @@
...
+  virtual std::vector<float> yolo_decode(float box_x, float box_y, float box_w, float box_h,
+                                         int stride, std::vector<std::vector<int>> anchor, int j,
+                                         int i, int a) const = 0;
+
+ private:
+  float score_thr_{0.4f};
+  int nms_pre_{1000};
+  float iou_threshold_{0.45f};
+  int min_bbox_size_{0};
+  std::vector<std::vector<std::vector<int>>> anchors_;
```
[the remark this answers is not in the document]
AllentDan: why would you say that?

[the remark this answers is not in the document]
AllentDan: `anchors_` here are all base anchors rather than `anchors_` after computation.

csrc/mmdeploy/codebase/mmdet/yolo_head.h
AllentDan: Seems CI failed with `using YOLOHead::YOLOHead;`. While local compiling passed.

lvhan028:
How about deriving another task from `class ObjectDetection` for rknn?

AllentDan:
> How about deriving another task from `class ObjectDetection` for rknn?

Why? Actually, I also used other backends with partition configs, which is convenient for debugging. These codes are more related to partition settings than just the RKNN backend.

mmdeploy/utils/config_utils.py
```diff
@@ -393,3 +393,51 @@ def get_dynamic_axes(
                 raise KeyError('No names were found to define dynamic axes.')
         dynamic_axes = dict(zip(axes_names, dynamic_axes))
     return dynamic_axes
+
+
+def get_normalization(model_cfg: Union[str, mmcv.Config]):
```
grimoire, line 398: docstring

mmdeploy/utils/config_utils.py
```diff
@@ -393,3 +393,51 @@ def get_dynamic_axes(
...
+def get_rknn_quantization(deploy_cfg: mmcv.Config):
+    """Get the flag of `do_quantization` for rknn backend.
+
+    Args:
+        deploy_cfg (mmcv.Config): The content of config.
+
+    Returns:
+        bool: Do quantization or not.
+    """
+    if get_backend(deploy_cfg) == Backend.RKNN:
```
grimoire, line 421: So we did not reuse calib_config?

mmdeploy/utils/config_utils.py
```diff
@@ -393,3 +393,51 @@ def get_dynamic_axes(
...
+        return get_backend_config(
+            deploy_cfg)['quantization_config']['do_quantization']
+    return False
+
+
+def disable_norm4rknn(transforms: Sequence[Dict], flag: bool = False):
+    """Set Normalization values for RKNN quantization.
+
+    Args:
+        common_params (Dict): Configs for rknn.config().
```
grimoire, line 431: No common_params

mmdeploy/codebase/mmaction/deploy/video_recognition.py
```diff
@@ -147,15 +147,17 @@ def init_pytorch_model(self,
 
     def create_input(self,
                      imgs: Union[str, np.ndarray, Sequence],
-                     input_shape: Sequence[int] = None) \
-            -> Tuple[Dict, torch.Tensor]:
+                     input_shape: Optional[Sequence[int]] = None,
+                     backend: Optional[Backend] = None,
```
grimoire, line 151: Why don't we move process_model_config into the task class to avoid add another parameter?

mmdeploy/utils/config_utils.py
AllentDan: Yes, quantization is controlled by a flag in RKNN.

mmdeploy/codebase/mmaction/deploy/video_recognition.py
AllentDan: Yes, it can be moved into task. But the parameter `backend` is to make difference between `creat_input`s of visualize pytorch model and backend model.

grimoire, line 151: Why do we add another parameter? task processor known the backend itself.

mmdeploy/backend/rknn/onnx2rknn.py
```diff
@@ -40,6 +42,14 @@ def onnx2rknn(onnx_model: str,
     output_names = onnx_params.get('output_names', None)
     input_size_list = get_backend_config(deploy_cfg).get(
         'input_size_list', None)
+    # update norm value
```
lvhan028, line 45: In my opinion, it's better to put `get_rknn_quantization` and `get_normalization` in `onnx2rknn.py` rather than `mmdeploy.utils`, cuz they are specially designed for rknn deployment.

AllentDan: But we import these methods in codebases.

lvhan028:
> > How about deriving another task from `class ObjectDetection` for rknn?
> 
> Why? Actually, I also used other backends with partition configs, which is convenient for debugging. These codes are more related to partition settings than just the RKNN backend.

I found the API of `BaseTask.create_input` is changed to the following:
```
    @abstractmethod
    def create_input(self,
                     imgs: Union[str, np.ndarray, Sequence],
                     input_shape: Optional[Sequence[int]] = None,
                     backend: Optional[Backend] = None,
                     **kwargs) -> Tuple[Dict, torch.Tensor]:
```
Make `backend` as an argument of `create_input` kinda odd.
So I propose to derive another task to reimplement `create_input` instead of changing the base class API.
But I realized this proposal requires a derived task in every codebase if deploying it to rknn, which is bad.

csrc/mmdeploy/codebase/mmdet/yolo_head.h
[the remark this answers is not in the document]
lzhangzz: Sorry, my mistake. should be `vector<vector<vector<float>>>` since you are reading from it by casting the values to `float` every time.

csrc/mmdeploy/codebase/mmdet/yolo_head.h
lzhangzz: You can inherit the base class constructors to avoid writing a constructors just to forward the arguments.

mmdeploy/utils/config_utils.py
grimoire: How did you feed dataset?

mmdeploy/backend/rknn/onnx2rknn.py
grimoire: Is it safe to call function in backends from codebase?

mmdeploy/utils/config_utils.py
AllentDan: by a txt file containing paths of images.

AllentDan:
@grimoire Now, the process is moved to the base task. New arguments for `create_input` and `process_model_cfg` are avoided.

mmdeploy/codebase/base/task.py
```diff
@@ -139,18 +140,38 @@ def single_gpu_test(self,
         return self.codebase_class.single_gpu_test(model, data_loader, show,
                                                    out_dir, **kwargs)
 
+    def reset_model_cfg_norm(self):
+        """Reset the norm values in model_cfg for some backends.
+
+        For example, in rknn-int8 backend, Normalize is moved to rknn models.
+        The processes of visualization and dump-info need this func.
+        """
+        if get_rknn_quantization(self.deploy_cfg):
```
grimoire, line 149: What if the user perform backend and pytorch inference in the same thread?

AllentDan: Do you mean running the RKNN model through adb remote and pytorch model at the same time? If there is really such a user, `model_cfg` need to be reloaded after normalization set to None.

grimoire: I don't want to add a function for rknn only in BaseTask.
This method could be used to adjust other backends with preprocessing supported. It is possible for OpenVINO user to compare the result of both pytorch and backend.

grimoire: Rename the method to a more **general** name, so we can use it to update pipeline for other backends.
In the future.

grimoire (approved):

mmdeploy/utils/config_utils.py
```diff
@@ -393,3 +393,40 @@ def get_dynamic_axes(
                 raise KeyError('No names were found to define dynamic axes.')
         dynamic_axes = dict(zip(axes_names, dynamic_axes))
     return dynamic_axes
+
+
+def get_normalization(model_cfg: Union[str, mmcv.Config]):
+    """Get the Nomalize transform from model config.
```
lvhan028, line 399: typo: 'Normalize'

lvhan028 (approved):

PR #1294, opened 2022-11-03
[Refactor] better SDK registry

lzhangzz:
1. Customizable creator signature
2. Declaration & definition of registries are not limited to namespace `mmdeploy` anymore
3. Faster look-up
4. Supports getting all creators as a linear array without any cost (`ModelRegistry` is not needed anymore)
5. Simplified interface


csrc/mmdeploy/core/model.h
```diff
@@ -70,24 +60,22 @@ class MMDEPLOY_API Model {
   Result<void> Init(const void* buffer, size_t size);
 
   /**
-   * @brief Return a specified model's meta info
-   * @param name the name of a model in sdk model file
+   * @brief Return a the model's meta info
```
lzhangzz, line 63: typo

csrc/mmdeploy/core/registry.h
```diff
@@ -3,55 +3,98 @@
...
-struct get_return_type<EntryType, std::void_t<typename EntryType::type>> {
-  using type = typename EntryType::type;
+namespace _registry {
+
+using std::optional;
+using std::string_view;
+
+template <typename T, typename = void>
+struct _get_signature {
+  static_assert(!std::is_same_v<T, T>, "tag T is not associated with a signature");
```
lvhan028, line 45: Why not `static_assert(0, ..)` directly?

lzhangzz: `static_assert(!std::is_same_v<T, T>)` only fails when the template gets instantiated, whereas `static_assert(0)` always fail.

csrc/mmdeploy/core/registry.h
```diff
@@ -63,67 +106,123 @@ class MMDEPLOY_API Registry<void> {
...
-  };                                                                   \
-  REGISTER_MODULE(base_type, module_name##Creator);
+#define MMDEPLOY_REGISTER_CREATOR(tag, creator_type)                    \
+  static ::mmdeploy::Registerer MMDEPLOY_ANONYMOUS_VARIABLE(register_){ \
+      [creator = creator_type{}]() mutable { ::mmdeploy::gRegistry<tag>().Add(creator); }};
+
+#define MMDEPLOY_CREATOR_DESC(name, version) #name, version
+
+#define MMDEPLOY_REGISTER_FACTORY_FUNC(tag, creator_desc, func)                                  \
+  static ::mmdeploy::Registerer MMDEPLOY_ANONYMOUS_VARIABLE(register_){                          \
```
lvhan028, line 222: `register_` is a static variable. Is it necessary to make it globally identical?

lzhangzz: depending on whether the `__COUNTER__` macro is supported, `MMDEPLOY_ANONYMOUS_VARIABLE(register_)` is going to be expanded to something like
```
register_0
register_1
register_2
```
or
```
register_ ## __LINE__
```
so as long as you don't put multiple `MMDEPLOY_REGISTER_FACTORY_FUNC` in the same line, it will be fine.

grimoire (approved):
LGTM

csrc/mmdeploy/codebase/mmocr/dbnet.cpp
```diff
@@ -137,8 +137,8 @@ class DBHead : public MMOCR {
 
 REGISTER_CODEBASE_COMPONENT(MMOCR, DBHead);
 
-}  // namespace mmocr
+MMDEPLOY_DEFINE_REGISTRY(DbHeadImpl);
 
-MMDEPLOY_DEFINE_REGISTRY(mmocr::DbHeadImpl);
+}  // namespace mmocr
```
lvhan028, line 142: May change to `namespace mmdeploy::mmocr{}` for consistency

csrc/mmdeploy/codebase/common.h
```diff
@@ -45,36 +44,34 @@ class CodebaseCreator : public Creator<Module> {
...
 
-#define REGISTER_CODEBASE_COMPONENT(codebase, component_type)                                      \
-  class component_type##_##Creator : public Creator<codebase> {                                    \
-   public:                                                                                         \
-    const char* GetName() const override { return #component_type; }                               \
-    int GetVersion() const override { return 1; }                                                  \
-    ReturnType Create(const Value& config) override { return CreateTask(component_type(config)); } \
-  };                                                                                               \
-  REGISTER_MODULE(codebase, component_type##_##Creator)
+#define REGISTER_CODEBASE_COMPONENT(codebase, component_type)                             \
```
lvhan028, line 71: Should we use `MMDEPLOY_REGISTER_CODEBASE_COMPONENT`

csrc/mmdeploy/core/registry.cpp
```diff
@@ -2,53 +2,80 @@
...
     }
+    names_.insert(v, creator.name());
+    creators_.insert(j, &creator);
+    return true;
   }
-  return nullptr;
+
+  Span<Creator<void>*> Creators() { return creators_; }
+
+  std::vector<Creator<void>*> creators_;
```
lvhan028, line 56: Should we put the member variables in private part?

lvhan028: Why use a vector to cache the creators? Is it better than `std::map`?
Everytime a new creator is added, the vector is probably re-arranged, isn't it?

lzhangzz: > Should we put the member variables in private part?

It does not matter, `struct Registry<void>::Impl` is private and hidden anyway.

> Why use a vector to cache the creators? Is it better than `std::map`?

It's simply a implementation of a `flat_multimap`, benchmark on `std::multimap` and `flat_multimap` reveals that if you have less than 1000 items, insertion on the flat variant is faster. And `std::equal_range` is always faster than `std::multimap::equal_range` due to better locality.

lvhan028 (approved):

PR #1314, opened 2022-11-07
[FIX] set stream argument when using async memcpy

francis0407:
## Motivation

The ops in tensorrt plugin should use asynchronous CUDA calls with a specified stream.

## Modification

set stream argument when using async memcpy


grimoire (approved):
LGTM

PR #1321, opened 2022-11-08
Add YOLOv5 support for RV1126 device.

AllentDan:
https://github.com/open-mmlab/mmyolo/pull/262

Require #1280


mmdeploy/backend/rknn/wrapper.py
```diff
@@ -67,3 +67,22 @@ def forward(self, inputs: Dict[str,
     def __rknnnn_execute(self, inputs: Sequence[np.array]):
         """Run inference with RKNN."""
         return self.rknn.inference(inputs)
+
+    def output_to_list(
```
grimoire, line 71: Why don't we enforce forward return a dict.

mmdeploy/codebase/mmdet/deploy/object_detection.py
```diff
@@ -66,9 +66,12 @@ def process_model_config(model_cfg: Config,
...
     for i, transform in enumerate(pipeline):
         # for static exporting
-        if input_shape is not None and transform.type == 'Resize':
-            pipeline[i].keep_ratio = False
-            pipeline[i].scale = tuple(input_shape)
+        if input_shape is not None:
+            if transform.type == 'Resize':
+                pipeline[i].keep_ratio = False
+                pipeline[i].scale = tuple(input_shape)
+            if 'Resize' in transform.type:
```
grimoire, line 73: Use equal instead of `in` here. Only process those transforms you known.

mmdeploy/backend/rknn/wrapper.py
AllentDan: Returning a dictionary is meaningless for the RKNN backend since it returns a list instead of a dictionary.

grimoire: It is the expected behavior of the wrapper.

AllentDan: I can return a dictionary then. But the key would be meaningless for users and may cause confusion.

grimoire: Did rknn change the output order?

AllentDan: Nope. I would set keys to string of numbers.

grimoire: Use the output names in the onnx as the key.

AllentDan: RKNN model has its own output names that are not ONNX out names.

AllentDan: Or, we can align the behavior to SDK such as [torchscript](https://github.com/open-mmlab/mmdeploy/blob/522fcc06350d426df0a152160b795507ad7982fc/csrc/mmdeploy/net/torchscript/torch_net.cpp#L126)

mmdeploy/apis/visualize.py
```diff
@@ -65,7 +65,10 @@ def visualize_model(model_cfg: Union[str, mmengine.Config],
         if backend == Backend.PYTORCH:
             model = task_processor.build_pytorch_model(model[0])
         else:
-            model = task_processor.build_backend_model(model)
+            model = task_processor.build_backend_model(
```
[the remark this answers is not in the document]
lvhan028: @AllentDan this logic is different from the master branch. what's the concern?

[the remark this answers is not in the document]
AllentDan: 1.x uses `data_preprocessor` to do the `Normalization`.

lvhan028:
@AllentDan may resolve the conflicts based on the latest dev-1.x

mmdeploy/backend/rknn/onnx2rknn.py
```diff
@@ -4,15 +4,28 @@
...
+    if toolkit is None:
+        return dict(name=None, version=None)
+    else:
+        return dict(name=toolkit.project_name, version=toolkit.version)
+
+
 def onnx2rknn(onnx_model: str,
               output_path: str,
               deploy_cfg: Union[str, mmengine.Config],
+              model_cfg: Union[str, mmengine.Config],
```
grimoire, line 28: Backend should not rely on model_cfg.
This is just a `TODO NOTE`, we can refactor it in the future.

mmdeploy/apis/visualize.py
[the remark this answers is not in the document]
grimoire: honestly, I kinda regret this design.

mmdeploy/backend/rknn/onnx2rknn.py
AllentDan: The same here. We can just pass normalization.

mmdeploy/codebase/mmdet/deploy/object_detection.py
```diff
@@ -262,14 +265,19 @@ def get_preprocess(self, *args, **kwargs) -> Dict:
             and 'Annotation' not in item['type']
         ]
         for i, transform in enumerate(transforms):
-            if transform['type'] == 'PackDetInputs':
+            if 'PackDetInputs' in transform['type']:
```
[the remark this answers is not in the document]
AllentDan: We don't remove `MMDet.` but replace all `Pack**Inputs` with Collect. So, I guess `in` should be okay here.

mmdeploy/apis/visualize.py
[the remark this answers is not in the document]
AllentDan: That may depend on what extent that we support other codebases. To ONNX, to Backends, Visualization and Benchmark testing are totally different.

grimoire (approved):

csrc/mmdeploy/preprocess/transform/letter_resize.cpp
```diff
@@ -0,0 +1,149 @@
...
+  explicit LetterResize(const Value& args) {
+    keep_ratio_ = args.value<bool>("keep_ratio", false);
+    if (args.contains("scale")) {
+      if (args["scale"].is_number_integer()) {
+        auto size = args["scale"].get<int>();
+        img_scale_ = {size, size};
+      } else if (args["scale"].is_array()) {
+        if (args["scale"].size() != 2) {
+          MMDEPLOY_ERROR("'scale' expects an array of size 2, but got {}", args["scale"].size());
+          throw std::length_error("'scale' expects an array of size 2");
```
lvhan028, line 28: Let's use `throw_exception(eInvalidArgument);` instead of `throw std::length_error`

csrc/mmdeploy/preprocess/transform/letter_resize.cpp
```diff
@@ -0,0 +1,149 @@
...
+        }
+        auto height = args["scale"][0].get<int>();
+        auto width = args["scale"][1].get<int>();
+        img_scale_ = {height, width};
+      } else {
+        MMDEPLOY_ERROR("'scale' is expected to be an integer or and array of size 2");
+        throw std::domain_error("'scale' is expected to be an integer or and array of size 2");
```
lvhan028, line 35: Let's use `throw_exception(eInvalidArgument);` instead of `throw std::domain_error`

csrc/mmdeploy/preprocess/transform/letter_resize.cpp
```diff
@@ -0,0 +1,149 @@
...
+    interpolation_ = args.value<string>("interpolation", "bilinear");
+    allow_scale_up_ = args.value<bool>("allow_scale_up", true);
+    use_mini_pad_ = args.value<bool>("use_mini_pad", false);
+    stretch_only_ = args.value<bool>("stretch_only", false);
+
+    vector<string> interpolations{"nearest", "bilinear", "bicubic", "area", "lanczos"};
+    if (std::find(interpolations.begin(), interpolations.end(), interpolation_) ==
+        interpolations.end()) {
+      MMDEPLOY_ERROR("'{}' interpolation is not supported", interpolation_);
+      throw std::invalid_argument("unexpected interpolation");
```
lvhan028, line 54: Let's use throw_exception(eInvalidArgument); instead of `throw std::invalid_argument`

configs/_base_/backends/rknn.py
```diff
@@ -1,8 +1,10 @@
...
-        mean_values=None,  # [[103.53, 116.28, 123.675]],
-        std_values=None,  # [[57.375, 57.12, 58.395]],
         target_platform='rv1126',  # 'rk3588'
         optimization_level=1),
-    quantization_config=dict(do_quantization=False, dataset=None))
+    quantization_config=dict(
+        do_quantization=True,
+        dataset=None,
+        pre_compile=False,
+        rknn_batch_size=-1))
```
lvhan028, line 10: why `-1`?

AllentDan: -1 is the default value of rknn-toolkit. It is the same with value 1.

mmdeploy/apis/visualize.py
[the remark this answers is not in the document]
lvhan028: This work around has been added to feishu projects for further discussed and addressed.
https://project.feishu.cn/openmmlab/story/detail/3227418

PR #1324, opened 2022-11-08
checkpoint load RuntimeError

MambaWong:
Checkpoint loading error when 'device=cpu'.
> RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu') to map your storages to the CPU.

### DIFF
--- a/mmdeploy/codebase/base/task.py
+++ b/mmdeploy/codebase/base/task.py
@@ -105,7 +105,10 @@ class BaseTask(metaclass=ABCMeta):
         model = MODELS.build(model)
         if model_checkpoint is not None:
             from mmengine.runner.checkpoint import load_checkpoint
-            load_checkpoint(model, model_checkpoint)
+            map_location = None
+            if self.device == 'cpu':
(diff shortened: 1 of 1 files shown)


mmdeploy/codebase/base/task.py
```diff
@@ -105,7 +105,10 @@ def build_pytorch_model(self,
         model = MODELS.build(model)
         if model_checkpoint is not None:
             from mmengine.runner.checkpoint import load_checkpoint
-            load_checkpoint(model, model_checkpoint)
+            map_location = None
+            if self.device == 'cpu':
+                map_location = torch.device('cpu')
+            load_checkpoint(model, model_checkpoint, map_location=map_location)
```
RunningLeon, line 111: hi, could change to `map_location=self.device`

grimoire (approved):
LGTM

RunningLeon (approved):
LGTM

PR #1325, opened 2022-11-08
checkpoint load RuntimeError

MambaWong:
Checkpoint loading error when 'device=cpu'.
> RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu') to map your storages to the CPU.


MambaWong:
网络卡，多刷新了几次，导致重复提了。

PR #1336, opened 2022-11-09
[Feature] Add resource monitor

irexyc:
## Motivation

add cpu/memory/device usage monitor

usage:
python tools/monitor.py cuda:0 python xxx
python tools/monitor.py cuda:0 ./bin xxx

### DIFF
--- a/tools/monitor.py
+++ b/tools/monitor.py
@@ -1,0 +1,185 @@
+import os
+import sys
+import subprocess
+import argparse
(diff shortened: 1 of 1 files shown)


AllentDan:
`rss` might raise error:
```
  File "tools/monitor.py", line 126, in show
    hwm = max(rss) >> 20
ValueError: max() arg is an empty sequence
```

PR #1363, opened 2022-11-14
add rknn device check

AllentDan:


mmdeploy/backend/rknn/__init__.py
```diff
@@ -13,6 +13,14 @@ def is_available():
     return importlib.util.find_spec('rknn') is not None
 
 
+def package_info():
+    import pkg_resources
+    for p in pkg_resources.working_set:
```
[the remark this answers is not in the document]
AllentDan: `rknn-toolkit` and `rknn-toolkit2` are different packages while share the same name `rknn` when importing. Already changed the way of search to speed up.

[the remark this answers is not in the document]
grimoire:
```python
toolkit = pkg_resources.working_set.by_key.get('rknn-toolkit', None)
toolkit = pkg_resources.working_set.by_key.get('rknn-toolkit2', toolkit)
```

hash is faster than loop string compare.

mmdeploy/backend/rknn/__init__.py
```diff
@@ -13,15 +13,26 @@ def is_available():
...
     """Check whether device available.
 
     Returns:
         bool: True if the device is available.
     """
-    ret = subprocess.check_output('adb devices', shell=True)
-    match = re.search(r'\\n\w+\\tdevice', str(ret))
-    return match is not None
+    try:
+        ret = subprocess.check_output('adb devices', shell=True)
```
grimoire, line 31: what if the connected device is not a rockchip device?

AllentDan: It's a behavior controlled by user. Any idea to improve the usage?

grimoire: I think device check is not necessary.

AllentDan: Okay, will delete the log info of adb device, which is a bonus actually. The function was designed to skip UT in CI since CI can install rknn-toolkit but could not connect a device.

grimoire: I don't think we should keep this function. As @lvhan028 mentioned, the device might still be available even when `device_available()==False`.

grimoire (approved):

lvhan028 (approved):

PR #1370, opened 2022-11-16
[FIX] Fix csharp net48 and batch inference

irexyc:
## Motivation

1. Support  .NET Framework 4.8
2. fix https://github.com/open-mmlab/mmdeploy/issues/1441

### DIFF
--- a/csrc/mmdeploy/apis/csharp/MMDeploy/NativeAPIs/NativeMethods.cs
+++ b/csrc/mmdeploy/apis/csharp/MMDeploy/NativeAPIs/NativeMethods.cs
@@ -5,6 +5,6 @@
     /// </summary>
     internal static partial class NativeMethods
     {
-        public const string DllExtern = "mmdeploy";
+        public const string DllExtern = "mmdeploy_extern";
     }
 }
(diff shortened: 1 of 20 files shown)


csrc/mmdeploy/apis/csharp/MMDeploy/NativeAPIs/NativeMethods.cs
```diff
@@ -5,6 +5,6 @@
     /// </summary>
     internal static partial class NativeMethods
     {
-        public const string DllExtern = "mmdeploy";
+        public const string DllExtern = "mmdeploy_extern";
```
lvhan028, line 8: Do we have to make another dll? 
If so, how about `mmdeploy_csharp`? `mmdeploy_extern` sounds ambiguous.

demo/csharp/pose_detection/pose_detection.csproj
```diff
@@ -14,7 +14,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="MMDeploy" Version="1.0.0" />
+    <PackageReference Include="MMDeployCSharp" Version="1.0.0" />
```
lvhan028, line 17: Let's use mmdeploy's version instead of "`1.0.0"

lvhan028 (approved):

lzhangzz (approved):
LGTM

PR #1371, opened 2022-11-16
[Feature] Support nvjpeg decoder

irexyc:
## Motivation

support jpeg device decoder

### DIFF
--- a/csrc/mmdeploy/apis/c/mmdeploy/image_decoder.cpp
+++ b/csrc/mmdeploy/apis/c/mmdeploy/image_decoder.cpp
@@ -129,7 +129,7 @@ int mmdeploy_image_decoder_apply(mmdeploy_image_decoder_t decoder, const char** 
   } catch (...) {
     MMDEPLOY_ERROR("unknown exception caught");
   }
-  MMDEPLOY_E_FAIL;
+  return MMDEPLOY_E_FAIL;
 }
 
 void mmdeploy_image_decoder_release_result(mmdeploy_mat_t* results, int count) {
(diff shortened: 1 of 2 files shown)


csrc/mmdeploy/apis/c/mmdeploy/image_decoder.h
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) OpenMMLab. All rights reserved.
+
+/**
+ * @file image_decoder.h
+ * @brief
```
lvhan028, line 5: Please make the brief clear

demo/csrc/c/benchmark_jpeg_decode.cpp
```diff
@@ -0,0 +1,85 @@
+#include <opencv2/core/types_c.h>
```
lvhan028, line 1: Examples in the demo folder aim at showing users how to use the API. It's better to make an example as simple as possible.
So, I suggest opening another cpp file, 'jpeg_decoder.cpp', presenting how to use the APIs declared in `image_decoder.h`

demo/csrc/c/benchmark_jpeg_decode.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include <string>
+#include <vector>
+
+#include "mmdeploy/image_decoder.h"
+
+using namespace std;
+
+int main(int argc, char **argv) {
+  if (argc != 4) {
+    printf("./benchmark_image_decode image_folder batch_size total\n");
```
lvhan028, line 19: miss returning error code

PR #1372, opened 2022-11-16
[Fix] fix regression test

AllentDan:


tests/regression/mmocr.yml
```diff
@@ -12,6 +12,7 @@ globals:
     word_acc:
       metric_key: IIIT5K/recog/word_acc_ignore_case_symbol
       tolerance: 0.05 # metric ±n%
+      dataset: IIIT5K
```
RunningLeon, line 15: do we need to add dataset for `hmean-iou` as well?

AllentDan: There are no concat datasets in the configs of the text-detection task.

RunningLeon (approved):
LGTM

PR #1379, opened 2022-11-17
[Feature] Support probability output for segmentation

AllentDan:


csrc/mmdeploy/codebase/mmseg/segment.cpp
```diff
@@ -31,10 +32,15 @@ class ResizeMask : public MMSegmentation {
     auto mask = inference_result["output"].get<Tensor>();
     MMDEPLOY_DEBUG("tensor.name: {}, tensor.shape: {}, tensor.data_type: {}", mask.name(),
                    mask.shape(), mask.data_type());
-    if (!(mask.shape().size() == 4 && mask.shape(0) == 1 && mask.shape(1) == 1)) {
+    if (!(mask.shape().size() == 4 && mask.shape(0) == 1)) {
```
RunningLeon, line 35: so here do not support batch > 1? this is not good for tensorrt where multi-batch could be used.

AllentDan: This is not introduced by this PR.

RunningLeon:
@AllentDan may fix conflicts

csrc/mmdeploy/codebase/mmseg/segment.cpp
RunningLeon: @lvhan028 Hi, not so sure about this. Does this mean no multiple batch inference is supported in skd for mmseg?

davodogster: Hi @RunningLeon I'm also very interested is this - Is batch inference possible? (n_samples > 1). 
For example, I have a script for test time augmentation which converts the image into 8 different images, and I want to process them all in one batch for inference. I am currently using:
image

Then I want to move this inference pipeline onto AWS for large scale GPU accelerated TRT batch inference.

Also, will TensorRT Segmentation probabilities (instead of class labels) be possible in the near future?
Thanks, Sam

lvhan028: > 

In SDK, batch inference tensors are carefully handled by `NetModule`. They will be sliced into single ones.
Refer to https://github.com/open-mmlab/mmdeploy/blob/b521e7da03c04fa5c783cb881f0f249f2cba9bb9/csrc/mmdeploy/net/net_module.cpp#L174
Thus, the postprocessing module just needs to focus on the `batch=1` case

davodogster: @lvhan028 Ok thanks, so Python API and model converter cannot do batch inference, but SDK can?

image

davodogster: @lvhan028 is SDK the only way to do batch inference for Semantic Segmentation in MMDeploy?

Also, is batch inference still possible if I generated the TRT engine model using?:
image

Thanks, Sam

RunningLeon (approved):
LGTM

configs/mmseg/segmentation_rknn-int8_static-320x320.py
```diff
@@ -2,6 +2,6 @@
 
 onnx_config = dict(input_shape=[320, 320])
 
-codebase_config = dict(model_type='rknn')
+codebase_config = dict(model_type='rknn', do_argmax=False)
```
lvhan028, line 5: Like other codebases, I think `with_argmax` is better than `do_argmax`

csrc/mmdeploy/codebase/mmseg/segment.cpp
```diff
@@ -44,6 +51,29 @@ class ResizeMask : public MMSegmentation {
...
+      int stride = height * width;
+      Tensor mask_out = TensorDesc{Device("cpu"),
+                                   DataType::kFLOAT,
+                                   {mask.shape(0), 1, mask.shape(2), mask.shape(3)},
+                                   "argmax_out"};
+      auto ptr = host_tensor.data<float>();
+      auto out_ptr = mask_out.data<float>();
+      for (int i = 0; i < stride; i++, ptr++) {
+        auto v = *ptr;
+        auto idx = 0.f;
```
lvhan028, line 65: Isn't `idx` an integer?

csrc/mmdeploy/codebase/mmseg/segment.cpp
```diff
@@ -44,6 +51,29 @@ class ResizeMask : public MMSegmentation {
     OUTCOME_TRY(auto host_tensor, MakeAvailableOnDevice(mask, host, stream_));
     OUTCOME_TRY(stream_.Wait());
 
+    if (!with_argmax_ && mask.shape(1) > 1 && mask.shape(1) == classes_ &&
+        host_tensor.data_type() == DataType::kFLOAT) {
+      int stride = height * width;
+      Tensor mask_out = TensorDesc{Device("cpu"),
```
lvhan028, line 57: why not output `masdk_out` directly?

lvhan028: `struct mmdeploy_segmentation_t` in `segmentor.h` can be defined as,
```c
typedef struct mmdeploy_segmentation_t {
  int height;   ///< height of \p mask that equals to the input image's height
  int width;    ///< width of \p mask that equals to the input image's width
  int classes;  ///< the number of labels in \p mask
  mmdeploy_data_type_t data_type; 
  void* mask;    ///< segmentation mask of the input image, in which mask[i * width + j] indicates
                ///< the label id of pixel at (i, j) if `data_type` is INT32. Otherwisse, mask means the probability feature map with shape as (classes, height, width)
} mmdeploy_segmentation_t;
```

csrc/mmdeploy/codebase/mmseg/segment.cpp
AllentDan: Yes， I tried to make it an int value but found that I have to make do `host_tensor = mask_out`. Or the program would crash when drawing masks in the demo.

csrc/mmdeploy/codebase/mmseg/segment.cpp
AllentDan: > why not output `masdk_out` directly? Let users decide how to do with the probability feature map

That would change other language interfaces like JAVA. I'm not familiar with that though.

lvhan028: I see.

csrc/mmdeploy/codebase/mmseg/segment.cpp
```diff
@@ -18,6 +18,8 @@ class ResizeMask : public MMSegmentation {
   explicit ResizeMask(const Value &cfg) : MMSegmentation(cfg) {
     try {
       classes_ = cfg["params"]["num_classes"].get<int>();
+      if (cfg["params"].contains("with_argmax"))
```
lvhan028, line 21: May use:
```c++
with_argmax_ = cfg["params"].value("with_argmax", true);
```

csrc/mmdeploy/codebase/mmseg/segment.cpp
lvhan028: I think the API has to be changed. A [feishu task](https://project.feishu.cn/openmmlab/story/detail/3325854) has been scheduled.
In this PR, I think logging an error message when `if (!with_argmax_)` is more appropriate.
```c++
if (!with_argmax_) {
    MMDEPLOY_ERROR("TODO: SDK will support probability featmap soon.");
    return Status(eNotSupported);
}
```

lvhan028 (approved):

PR #1391, opened 2022-11-18
[FIX] Fix sdk export pipeline.json

irexyc:
## Motivation

Fix sdk export pipeline.json. 

The error is due to line 249

### DIFF
--- a/mmdeploy/backend/sdk/export_info.py
+++ b/mmdeploy/backend/sdk/export_info.py
@@ -245,8 +245,9 @@ def get_pipeline(deploy_cfg: mmengine.Config, model_cfg: mmengine.Config,
     task = get_task_type(deploy_cfg)
     input_names = preprocess['input']
     output_names = postprocess['output']
-    if task == Task.CLASSIFICATION or task == Task.SUPER_RESOLUTION \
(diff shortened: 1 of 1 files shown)


mmdeploy/backend/sdk/export_info.py
```diff
@@ -245,8 +245,9 @@ def get_pipeline(deploy_cfg: mmengine.Config, model_cfg: mmengine.Config,
     task = get_task_type(deploy_cfg)
     input_names = preprocess['input']
     output_names = postprocess['output']
-    if task == Task.CLASSIFICATION or task == Task.SUPER_RESOLUTION \
-            or Task.VIDEO_RECOGNITION:
+    if task in [
```
RunningLeon, line 248: why need to change here? the logic seems unchange.

irexyc: It should be `task == Task.VIDEO_RECOGNITION` not just `Task.VIDEO_RECOGNITION`

RunningLeon: ooops

RunningLeon (approved):
LGTM

PR #1403, opened 2022-11-21
[Enhancement] refactor profiler

irexyc:
## Motivation

reduce profile node name

### DIFF
--- a/csrc/mmdeploy/graph/inference.cpp
+++ b/csrc/mmdeploy/graph/inference.cpp
@@ -35,9 +35,12 @@ Result<unique_ptr<Node>> InferenceBuilder::BuildImpl() {
 
   auto pipeline_config = from_json<Value>(json);
   if (context.contains("scope")) {
-    auto net_cfg = pipeline_config["pipeline"]["tasks"][1];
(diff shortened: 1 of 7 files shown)


csrc/mmdeploy/graph/inference.cpp
```diff
@@ -35,9 +35,12 @@ Result<unique_ptr<Node>> InferenceBuilder::BuildImpl() {
 
   auto pipeline_config = from_json<Value>(json);
   if (context.contains("scope")) {
-    auto net_cfg = pipeline_config["pipeline"]["tasks"][1];
-    std::string net_name = net_cfg["name"].get<std::string>();
-    auto scope = context["scope"].get_ref<profiler::Scope*&>()->CreateScope(net_name);
+    auto name = config_.value("name", std::string());
+    if (name == "") {
```
lzhangzz, line 39: use something like 
```C++
auto name = config_.value("name", config_["type"].get<std::string>());
```
fallback mechanism should not be built on assumptions of pipeline structure.

lzhangzz (approved):
LGTM

lvhan028 (approved):

PR #1410, opened 2022-11-22
[Fix] fix mmaction2 docs

irexyc:
## Motivation

fix type and add link to README.md


RunningLeon:
@irexyc seems it could be put in pr #1411

RunningLeon:
@irexyc en docs should be updated as well https://github.com/open-mmlab/mmdeploy/blob/master/docs/en/04-supported-codebases/mmaction2.md. `mmocr` appears in the doc.

AllentDan (approved):

RunningLeon (approved):
LGTM

PR #1420, opened 2022-11-23
[Fix] fix cls head in SDK

AllentDan:
when `topk` is greater than `class_num`. Appeared in customized datasets.

### DIFF
--- a/csrc/mmdeploy/codebase/mmcls/linear_cls.cpp
+++ b/csrc/mmdeploy/codebase/mmcls/linear_cls.cpp
@@ -37,7 +37,6 @@ class LinearClsHead : public MMClassification {
     }
 
     auto class_num = (int)output.shape(1);
-    topk_ = topk_ < class_num ? topk_ : class_num;
 
     OUTCOME_TRY(auto _scores, MakeAvailableOnDevice(output, kHost, stream()));
     OUTCOME_TRY(stream().Wait());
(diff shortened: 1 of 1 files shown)


irexyc:
may also add the change to  [multi_label_linear_cls.cpp](https://github.com/open-mmlab/mmdeploy/blob/0d6868a35b346e555b17375bab1632857cf76794/csrc/mmdeploy/codebase/mmcls/multi_label_linear_cls.cpp)

AllentDan:
> may also add the change to [multi_label_linear_cls.cpp](https://github.com/open-mmlab/mmdeploy/blob/0d6868a35b346e555b17375bab1632857cf76794/csrc/mmdeploy/codebase/mmcls/multi_label_linear_cls.cpp)

Hi, @irexyc . What's the potential risk for multi_label? I didn't see `topk` there.

irexyc (approved):

csrc/mmdeploy/codebase/mmcls/linear_cls.cpp
```diff
@@ -37,6 +37,7 @@ class LinearClsHead : public MMClassification {
     }
 
     auto class_num = (int)output.shape(1);
+    topk_ = topk_ < class_num ? topk_ : class_num;
```
lzhangzz, line 40: Modifying class member `top_k_` makes the module stateful, which is hard to reason locally.

Add `auto top_k = std::min(top_k_, class_num);` in `GetLabels` and use it as loop condition instead of modifying `top_k_`

lzhangzz (approved):

PR #1424, opened 2022-11-23
[Fix] fix visualization for partition

AllentDan:
Introduced by #1280

### DIFF
--- a/mmdeploy/codebase/base/backend_model.py
+++ b/mmdeploy/codebase/base/backend_model.py
@@ -6,8 +6,7 @@ import mmcv
 import torch
 
 from mmdeploy.utils import (SDK_TASK_MAP, Backend, get_backend_config,
-                            get_common_config, get_ir_config,
-                            get_partition_config, get_task_type)
+                            get_common_config, get_ir_config, get_task_type)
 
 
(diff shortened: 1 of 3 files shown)


mmdeploy/codebase/base/backend_model.py
```diff
@@ -26,6 +27,9 @@ def __init__(self,
         if deploy_cfg is not None:
             ir_config = get_ir_config(deploy_cfg)
             output_names = ir_config.get('output_names', None)
+            if get_partition_config(deploy_cfg) is not None:
+                output_names = get_partition_config(
+                    deploy_cfg)['partition_cfg'][0]['output_names']
```
grimoire, line 32: what if the model will be partition into two part?

AllentDan: `output_names` should be passed into the class instead of `None`.

grimoire: two stage model will be partitioned into two part, which output_names will you accept?

grimoire (approved):

RunningLeon (approved):
LGTM

PR #1430, opened 2022-11-24
[Fix] fix ncnn torch 1.12 master

hanrui1sensetime:

## Motivation

In torch 1.12, some constant initializer which uses in multiple places will generate `Identity` ops in onnx, which will failed in ncnn.
So we fix the fuse pass in the onnx2ncnn to remove these `Identity` nodes.

## Modification

Fuse_pass.cpp and onnx2ncnn.cpp

## BC-breaking (Optional)

None.

## Use cases (Optional)

In torch1.12 ncnn deployment which failed before. A typical example are models which have some `Resize` ops which reuse `scale` initializers.

## Checklist

### DIFF
--- a/csrc/mmdeploy/backend_ops/ncnn/onnx2ncnn/fuse_pass.cpp
+++ b/csrc/mmdeploy/backend_ops/ncnn/onnx2ncnn/fuse_pass.cpp
@@ -1,6 +1,31 @@
 // Copyright (c) OpenMMLab. All rights reserved.
 #include "fuse_pass.h"
 
+void fuse_identity(onnx::GraphProto* mutable_graph,
+                   std::map<std::string, onnx::TensorProto>& weights,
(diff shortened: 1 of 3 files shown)


tpoisonooo (approved):
LGTM after  https://github.com/open-mmlab/mmdeploy/pull/1431 merged.

grimoire:
Why do we write a pass? we can just map Identity to Noop

hanrui1sensetime:
> Why do we write a pass? we can just map Identity to Noop

Noop requires one bottomtop blob, while Identity requires one bottom blob and one top blob. If we map Identity to Noop, we still need substitute the Identity output node to Identity input node for nodes which using Identity's output as input. This step is the same as the fuse_identity function.

RunningLeon (approved):
LGTM

PR #1484, opened 2022-12-04
[Improve] Improve dockerfile pip source

PeterH0323:
## Motivation

When using `USE_SRC_INSIDE=true` the pip source still the original. Since the user want to use `USE_SRC_INSIDE` so we need to switch the source of pip as well

## Modification

Add pip source. And speed up the download for pip . 🚀


AllentDan:
Hi, the same to the CPU dockerfile please.

RunningLeon (approved):
LGTM

AllentDan (approved):
LGTM

lvhan028:
@PeterH0323 Could you please pull another PR to the master branch?

PR #1492, opened 2022-12-06
[Enhancement] Add pip source in dockerfile for `master` branch

PeterH0323:
## Motivation

When using `USE_SRC_INSIDE=true` the pip source still the original. Since the user want to use `USE_SRC_INSIDE` so we need to switch the source of pip as well

## Modification

Add pip source. And speed up the download for pip . 🚀 

Is a copy PR #1484 for master branch


AllentDan (approved):

RunningLeon (approved):
LGTM

PR #1509, opened 2022-12-09
[Fix] Remove cudnn dependency for transform 'mmaction2::format_shape'

irexyc:
## Motivation

1. remove cudnn deps for mmaction2::format_shape
2. fix [FormatShape](csrc/mmdeploy/codebase/mmaction/format_shape.cpp), it accept a list but return an object.
3. FormatShape forget to init op.


lzhangzz (approved):

csrc/mmdeploy/codebase/mmaction/format_shape.cpp
```diff
@@ -50,6 +51,7 @@ Result<void> FormatShape::Apply(Value& data) {
   }
 
   Tensor dst;
+  data = Value{};
```
lvhan028, line 54: why clear `data`?

irexyc: The next node treate data as a object, while in this node, the data is a list.

lvhan028 (approved):

PR #1514, opened 2022-12-10
[Refactor] Support custom codebase and task

grimoire:

## Motivation

Ease adding custom codebase.

## Modification

- custom import
- Codebase registry will add unregistered codebase to enum automatically.
- Alias MMCodebase -> BaseCodebase
- Rename xxxx-> xxxxTask

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdeploy/apis/utils/utils.py
+++ b/mmdeploy/apis/utils/utils.py
@@ -37,10 +37,10 @@ def build_task_processor(model_cfg: mmengine.Config,
         BaseTask: A task processor.
     """
     check_backend_device(deploy_cfg=deploy_cfg, device=device)
-    codebase_type = get_codebase(deploy_cfg)
+    codebase_name = get_codebase(deploy_cfg, return_enum=False)
     custom_module_list = get_codebase_external_module(deploy_cfg)
-    import_codebase(codebase_type, custom_module_list)
-    codebase = get_codebase_class(codebase_type)
+    import_codebase(codebase_name, custom_module_list)
+    codebase = get_codebase_class(codebase_name)
     return codebase.build_task_processor(model_cfg, deploy_cfg, device)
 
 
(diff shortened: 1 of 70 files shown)


tools/deploy.py
```diff
@@ -58,6 +59,11 @@ def parse_args():
         '--uri',
         default='192.168.1.1:60000',
         help='Remote ipv4:port or ipv6:port for inference on edge device.')
+    parser.add_argument(
```
RunningLeon, line 62: may update the docs

mmdeploy/apis/utils/utils.py
```diff
@@ -37,10 +37,11 @@ def build_task_processor(model_cfg: mmengine.Config,
         BaseTask: A task processor.
     """
     check_backend_device(deploy_cfg=deploy_cfg, device=device)
-    codebase_type = get_codebase(deploy_cfg)
-    custom_module_list = get_codebase_external_module(deploy_cfg)
-    import_codebase(codebase_type, custom_module_list)
-    codebase = get_codebase_class(codebase_type)
+    codebase_name = get_codebase(deploy_cfg, return_enum=False)
```
RunningLeon, line 40: Could fix conflicts

mmdeploy/codebase/__init__.py
```diff
@@ -22,17 +18,19 @@ def import_codebase(codebase_type: Codebase, custom_module_list: List = []):
         codebase (Codebase): The codebase to import.
```
hanrui1sensetime, line 18: May update the args docstring.

hanrui1sensetime:
Please fix conflict.

AllentDan:
Does this mean that only modules that formed in package and can be imported by `importlib` were supported? In many cases, users only write a model in a script using mmxxx.

grimoire:
> Does this mean that only modules that formed in package and can be imported by `importlib` were supported? In many cases, users only write a model in a script using mmxxx.

script in current workspace will be treated as module by python.

AllentDan:
> > Does this mean that only modules that formed in package and can be imported by `importlib` were supported? In many cases, users only write a model in a script using mmxxx.
> 
> script in current workspace will be treated as module by python.

I see.

Issue #1529, opened 2022-12-13
[Bug] When I use mmdeploy to convert a pytorch model to onnx model, how can I set the opset version?


labmanno004:
### Checklist

- [X] I have searched related issues but cannot get the expected help.
- [X] 2. I have read the [FAQ documentation](https://github.com/open-mmlab/mmdeploy/blob/master/docs/en/faq.md) but cannot get the expected help.
- [X] 3. The bug has not been fixed in the latest version.

### Describe the bug

I did not find the corresponding introduction in the document. 
The opset version of the model generated by default is v11, but I need v13.

### Reproduction

/

### Environment

```Shell
/
```

AllentDan:
There is an argument `opset_version` in `configs/_base_/onnx_config.py` for users.

labmanno004:
> There is an argument `opset_version` in `configs/_base_/onnx_config.py` for users.

OK. Thank you very much.

labmanno004:
> There is an argument `opset_version` in `configs/_base_/onnx_config.py` for users.

An error was reported after the opset_version was changed from 11 to 13. And when opset_version is 11, the code runs successfully.

root@sse-lg-112-41:~/workspace# python mmdeploy/tools/deploy.py mmdeploy/configs/mmocr/text-detection/text-detection_onnxruntime_dynamic.py mmocr/configs/textdet/dbnet/dbnet_resnet18_fpnc_1200e_icdar2015.py checkpoints/dbnet_resnet18_fpnc_1200e_icdar2015_20220825_221614-7c0e94f2.pth mmocr/demo/demo_text_det.jpg --work-dir mmdeploy_model/dbnet-op13 --device cpu --dump-info
[2022-12-13 07:43:42.578] [mmdeploy] [warning] [registry.cpp:44] Adding duplicated creator (ResizeOCR, 0).
[2022-12-13 07:43:45.399] [mmdeploy] [warning] [registry.cpp:44] Adding duplicated creator (ResizeOCR, 0).
12/13 07:43:47 - mmengine - WARNING - Failed to search registry with scope "mmocr" in the "Codebases" registry tree. As a workaround, the current "Codebases" registry in "mmdeploy" is used to build instance. This may cause unexpected failure when running the built modules. Please check whether "mmocr" is a correct scope, or whether the registry is initialized.
[...]

PR #1538, opened 2022-12-15
[Fix] fix squeeze export and unsqueeze pass for opset 13

AllentDan:
#1529

### DIFF
--- a/mmdeploy/pytorch/symbolics/squeeze.py
+++ b/mmdeploy/pytorch/symbolics/squeeze.py
@@ -1,7 +1,9 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+import torch
 import torch.onnx.symbolic_helper as sym_help
 
 from mmdeploy.core import SYMBOLIC_REWRITER
+from mmdeploy.utils.config_utils import get_ir_config
 
 
 @SYMBOLIC_REWRITER.register_symbolic('squeeze', is_pytorch=True)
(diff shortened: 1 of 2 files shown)


mmdeploy/pytorch/symbolics/squeeze.py
```diff
@@ -18,4 +20,10 @@ def squeeze__default(g, self, dim=None):
                 dims.append(i)
     else:
         dims = [sym_help._get_const(dim, 'i', 'dim')]
+
+    ctx = SYMBOLIC_REWRITER.get_context('squeeze')
+    if get_ir_config(ctx.cfg)['opset_version'] >= 13:
```
grimoire, line 25: opset_version can be get from [g.opset](https://github.com/pytorch/pytorch/blob/6a866c3ed180c05520e2114e48e97bf1e8b2e1a3/torch/onnx/symbolic_helper.py#L815). So we can export it without deploy config.

grimoire (approved):
Opset 11 and opset 13 has changed a lot.
It is risky to maintain multiple opset version support.

mmdeploy/pytorch/symbolics/squeeze.py
AllentDan: I tried to but failed because our `g` is an object of `torch._C.graph` instead of `jit_utils.GraphContext`.

RunningLeon (approved):
LGTM

Issue #1537, opened 2022-12-14
[Bug] [Windows] Error: "Expression: cannot seek vector iterator after end" when running mmdeploy::Detector::Apply in debug version of mmdeploy library.


antoszy:
### Checklist

- [X] I have searched related issues but cannot get the expected help.
- [X] 2. I have read the [FAQ documentation](https://github.com/open-mmlab/mmdeploy/blob/master/docs/en/faq.md) but cannot get the expected help.
- [X] 3. The bug has not been fixed in the latest version.

### Describe the bug

When running `mmdeploy::Detector::Apply` function I get the following error: 
![image](https://user-images.githubusercontent.com/23712573/207630379-b5fcb91c-24b3-44f9-ae8b-df355ce204b6.png)

### Reproduction

Build debug version of the mmdeploy libraries with:
```
cmake .. -G "Visual Studio 16 2019" -A x64 -T v142 `
   -DMMDEPLOY_BUILD_SDK=ON `
   -DMMDEPLOY_BUILD_EXAMPLES=ON `
   -DMMDEPLOY_BUILD_SDK_PYTHON_API=ON `
   -DMMDEPLOY_BUILD_SDK_CXX_API=ON `
   -DMMDEPLOY_TARGET_DEVICES="cuda" `
   -DMMDEPLOY_TARGET_BACKENDS="trt" `
   -Dpplcv_DIR="$env:PPLCV_DIR/build_d/install/lib/cmake/ppl" `
   -DTENSORRT_DIR="$env:TENSORRT_DIR" `
   -DCUDNN_DIR="$env:CUDNN_DIR" `
   -Wno-dev `
   -DOpenCV_DIR="$env:OPENCV_DEBUG_DIR\..\.."
cmake --build . --config Debug -- /m
cmake --install . --config Debug
```

Crate TensorRT version of the model by running the following command:
```
python mmdeploy\tools\deploy.py \
	.\mmdeploy\configs\mmdet\instance-seg\instance-seg_tensorrt_static-800x1344.py  \
	.\mmdetection\configs\mask_rcnn\mask_rcnn_r50_fpn_1x_coco.py  \
	.\mmmodels\mask_rcnn_r50_fpn_2x_coco_bbox_mAP-0.392__segm_mAP-0.354_20200505_003907-3e542a40.pth  \
	.\mmdetection\demo\demo.jpg  \
	--work-dir .\mmdeploy_model\mask-rcnn  \
	--device cuda  \
[...]

PR #1543, opened 2022-12-15
 Fix 'cannot seek vector iterator' in debug windows build

antoszy:
## Motivation

Fixes: #1537 

## Modification

Adding `nbInputs` to the `mFeatmapStrides.begin()` iterator cases it to go out of range of the `mFeatmapStrides` vector. When this is compiled in debug build in MS compiler, it causes an exception. If we decrement the value first and then add it to iterator, than the iterator stays in correct vector range.

## Checklist


lvhan028 (approved):
LGTM

irexyc (approved):

lvhan028:
@antoszy could you also open a PR to master branch?

PR #1546, opened 2022-12-15
[Fix] rename mmaction -> mmaction2 in mmdeploy namespace (dev-1.x)

irexyc:
## Motivation

rename mmaction -> mmaction2


mmdeploy/backend/sdk/export_info.py
```diff
@@ -32,7 +32,10 @@ def get_task(deploy_cfg: mmengine.Config) -> Dict:
     task_name = get_task_type(deploy_cfg).value
     codebase_name = get_codebase(deploy_cfg).value
     try:
-        codebase = importlib.import_module(codebase_name)
+        codebase_lib_name = codebase_name
```
lvhan028, line 35: I think this is not a good idea.
@grimoire and @tpoisonooo any suggestions?

grimoire: Can we add `get_version` method to classes derived from MMCodebase? It can also be used in check env.

RunningLeon:
@irexyc Could fix conflict and update this pr

PR #1552, opened 2022-12-16
[Fix] fix unittest and suppress warning

grimoire:

## Motivation

Fix unit test.

## Modification

- string cmp in `trt_roi_align.cpp`
- TensorRT 8.5 api update
- `test_core` bad rewriter error.
- `nms.py` wrong rewriter
- remove unsed test.
- det3d models test recovery error.
- suppress deprecate warning and user warning.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
+++ b/csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
@@ -203,9 +203,9 @@ nvinfer1::IPluginV2 *TRTRoIAlignCreator::createPlugin(
       int data_size = fc->fields[i].length;
       const char *data_start = static_cast<const char *>(fc->fields[i].data);
       std::string poolModeStr(data_start, data_size);
-      if (poolModeStr == "avg") {
+      if (strcmp(poolModeStr.c_str(), "avg") == 0) {
         poolMode = 1;
-      } else if (poolModeStr == "max") {
+      } else if (strcmp(poolModeStr.c_str(), "max") == 0) {
         poolMode = 0;
       } else {
         std::cout << "Unknown pool mode \"" << poolModeStr << "\"." << std::endl;
--- a/csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
+++ b/csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
@@ -203,13 +203,13 @@ nvinfer1::IPluginV2 *TRTRoIAlignCreator::createPlugin(
       int data_size = fc->fields[i].length;
       ASSERT(data_size > 0);
       const char *data_start = static_cast<const char *>(fc->fields[i].data);
-      std::string poolModeStr(data_start);
-      if (poolModeStr == "avg") {
+      std::string pool_mode(data_start);
+      if (pool_mode == "avg") {
         poolMode = 1;
-      } else if (poolModeStr == "max") {
+      } else if (pool_mode == "max") {
         poolMode = 0;
       } else {
-        std::cout << "Unknown pool mode \"" << poolModeStr << "\"." << std::endl;
+        std::cout << "Unknown pool mode \"" << pool_mode << "\"." << std::endl;
       }
       ASSERT(poolMode >= 0);
     }
--- a/csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
+++ b/csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
@@ -201,11 +201,12 @@ nvinfer1::IPluginV2 *TRTRoIAlignCreator::createPlugin(
 
     if (field_name.compare("mode") == 0) {
       int data_size = fc->fields[i].length;
+      ASSERT(data_size > 0);
       const char *data_start = static_cast<const char *>(fc->fields[i].data);
-      std::string poolModeStr(data_start, data_size);
-      if (strcmp(poolModeStr.c_str(), "avg") == 0) {
+      std::string poolModeStr(data_start);
+      if (poolModeStr == "avg") {
         poolMode = 1;
-      } else if (strcmp(poolModeStr.c_str(), "max") == 0) {
+      } else if (poolModeStr == "max") {
         poolMode = 0;
       } else {
         std::cout << "Unknown pool mode \"" << poolModeStr << "\"." << std::endl;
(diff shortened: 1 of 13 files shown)


csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
```diff
@@ -203,9 +203,9 @@ nvinfer1::IPluginV2 *TRTRoIAlignCreator::createPlugin(
       int data_size = fc->fields[i].length;
       const char *data_start = static_cast<const char *>(fc->fields[i].data);
       std::string poolModeStr(data_start, data_size);
-      if (poolModeStr == "avg") {
+      if (strcmp(poolModeStr.c_str(), "avg") == 0) {
```
lvhan028, line 206: operator '==' is done via `string::compare()`.
For checking whether two strings are the same (true or false) instead of comparing which is greater or not, I think `==` is fine.

mmdeploy/backend/tensorrt/utils.py
```diff
@@ -13,15 +13,17 @@
 from .init_plugins import load_tensorrt_plugin
 
 
-def save(engine: trt.ICudaEngine, path: str) -> None:
+def save(engine: Any, path: str) -> None:
```
lvhan028, line 16: Besides `trt.ICudaEngine` what else type might `engine` be?

lvhan028:
May resolve the linting issue

mmdeploy/codebase/mmdet/models/backbones.py
```diff
@@ -46,7 +46,7 @@ def focus__forward__ncnn(self, x):
 
     x = x.reshape(batch_size, c * h, 1, w)
     _b, _c, _h, _w = x.shape
-    g = _c // 2
+    g = torch.div(_c, 2, rounding_mode='floor')
```
lvhan028, line 49: What's the benefit?

csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
grimoire: poolModeStr is a std::string with data "avg\0"

mmdeploy/backend/tensorrt/utils.py
grimoire: IHostMemory, which is the output of new API in TensorRT 8.5+

mmdeploy/codebase/mmdet/models/backbones.py
grimoire: deprecate warning

mmdeploy/backend/tensorrt/utils.py
grimoire: https://docs.nvidia.com/deeplearning/tensorrt/api/python_api/infer/Core/Builder.html#tensorrt.Builder.build_serialized_network

mmdeploy/codebase/mmdet/models/backbones.py
AllentDan: Why not just rewrite `floordiv`?

csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
lvhan028: https://en.cppreference.com/w/cpp/string/basic_string/operator_cmp
> Compares the contents of a string with another string or a null-terminated array of CharT.

mmdeploy/codebase/mmdet/models/backbones.py
grimoire: Rewriter should not change the behavior of the function.

mmdeploy/backend/tensorrt/utils.py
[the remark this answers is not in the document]
grimoire: I am not sure if IHostMemory is available in trt7

csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
grimoire: `the contents of a string` contains a null byte. The comparison will always be failed (Yes, it did in the unit test).
If you really prefer the operator, I will cast the c-string to a std::string.

AllentDan (approved):

csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
lvhan028: Could you show me your UT?
here is what I did
```C++
#include <iostream>

#include <string>


int main() {
  std::string pool_mode("avg");
  std::cout << (pool_mode == "AVG") << ", " << (pool_mode == "avg") << std::endl;
  return 0;
}

```
The output is: 0, 1

grimoire: This is THE string.
```c++
char data[4] = {'a','b','c','\0'};
string pool_mode(pool_mode, 4);
```

lvhan028: Why not using `string pool_mode(data)`?

lvhan028: > This is THE string.
> 
> ```c++
> char data[4] = {'a','b','c','\0'};
> string pool_mode(pool_mode, 4);
> ```

This initialization indeed treats `pool_mode` as a binary buffer rather than a string

grimoire: Yes, it did.
https://github.com/open-mmlab/mmdeploy/blob/e76c0a8a8ba24b1254ef299d2efc6bf9feeb57b8/csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp#L203-L208

The length varies between different version, what can I do?

grimoire: > Why not use `string pool_mode(data)`?

https://github.com/open-mmlab/mmdeploy/blob/e76c0a8a8ba24b1254ef299d2efc6bf9feeb57b8/csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp#L203

The data is read from the serialized data. And the length varies between different versions. What if they decide to omit null bytes as long as they already provide the data length?

lvhan028 (approved):

PR #1605, opened 2023-01-04
[Enhancement] Add windows-cuda-build ci

irexyc:
## Motivation

Add windows-cuda-trt build check

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -243,8 +243,8 @@ jobs:
           echo "OPENCV_DIR=$pwd\opencv-4.5.5\x64\vc16\lib" >> $env:GITHUB_ENV
       - name: Install pplcv
         run: |
-          Invoke-WebRequest -Uri https://github.com/irexyc/mmdeploy-ci-resource/releases/download/pplcv/pplcv_cuda10.2.zip -OutFile pplcv_cuda10.2.zip
-          Expand-Archive pplcv_cuda10.2.zip pplcv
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -207,3 +207,143 @@ jobs:
           env_vars: OS,PYTHON,CPLUS
           name: codecov-umbrella
           fail_ci_if_error: false
+
+  build_cuda102-windows:
```
RunningLeon, line 211: the newly added ci failed.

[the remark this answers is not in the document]
irexyc: I think it's better not to combine since it takes  long time to build and two tasks can run in parallel.

irexyc: will be fixed by https://github.com/open-mmlab/mmdeploy/pull/1626

.github/workflows/build.yml
```diff
@@ -207,3 +207,143 @@ jobs:
...
+          Invoke-WebRequest -Uri https://github.com/irexyc/mmdeploy-ci-resource/releases/download/pplcv/pplcv-0.6.2.windows10.x86_64.cuda-10.2.zip -OutFile pplcv.zip
+          Expand-Archive pplcv.zip pplcv
+          echo "PPLCV_DIR=$pwd\pplcv\lib\cmake\ppl" >> $env:GITHUB_ENV
+      - name: Build
+        run: |
+          mkdir build
+          cd build
+          cmake ..  -A x64 -T v142 `
+            -DMMDEPLOY_SHARED_LIBS=OFF `
+            -DMMDEPLOY_BUILD_SDK_CXX_API=ON `
```
lvhan028, line 255: `MMDEPLOY_BUILD_SDK_CXX_API` option has been removed.

PR #1611, opened 2023-01-05
[Fix]: fix numba numpy version not compatiable in circileci

RunningLeon:
## Motivation

fix numba numpy version not compatiable in circileci

## Modification

- Reinstall latest numpy and numba in circleci workflow
- Fix mmcls ut
- Fix dead links in documents

## BC-breaking (Optional)

None

## Use cases (Optional)

## Checklist


.circleci/config.yml
```diff
@@ -25,12 +25,14 @@ workflows:
             configs/.* lint_only false
```
AllentDan, line 25: configs should be lint_only true?

RunningLeon: yes.

AllentDan (approved):
LGTM

.circleci/config.yml
```diff
@@ -22,15 +22,17 @@ workflows:
           mapping: |
             .circle/.* lint_only false
             cmake/.* lint_only false
-            configs/.* lint_only false
+            configs/.* lint_only true
```
lvhan028, line 25: May remove the `lint_only true` lines, since the default value of `lint_only` is true in `test.yml`

AllentDan: I found that the setting seemed not working, #1621 for example.

lvhan028: Is it because `config.yml` has been changed?

AllentDan: No matter I add `docs/.* lint_only true` into `config.yml` or not, circle CI run `test_linux_onnxruntime`. I only changed docs in that PR.

lvhan028: https://github.com/open-mmlab/mmocr/blob/27b6a68586b9a040678fe083bcf60662ae1b9261/.circleci/test.yml#L126
I found an example in mmocr:
```
pr_stage_lint:
    when: << pipeline.parameters.lint_only >>
    jobs:
      - lint:
          name: lint
          filters:
            branches:
              ignore:
                - dev-1.x
                - 1.x
```
I think we should also try the `filters` option

RunningLeon: > https://github.com/open-mmlab/mmocr/blob/27b6a68586b9a040678fe083bcf60662ae1b9261/.circleci/test.yml#L126 I found an example in mmocr:
> 
> ```
> pr_stage_lint:
>     when: << pipeline.parameters.lint_only >>
>     jobs:
>       - lint:
>           name: lint
>           filters:
>             branches:
>               ignore:
>                 - dev-1.x
>                 - 1.x
> ```
> 
> I think we should also try the `filters` option

It's a bit confusing. If `pipeline.parameters.lint_only` is `True`, then we should do lint check and do not care about the branch.

RunningLeon: @lvhan028 @AllentDan maybe #1621 is because of not changing the `base-revision` in circleci config
https://github.com/open-mmlab/mmdeploy/pull/1621/commits/b0c5513f533ae7d6aac4668549e770d722a4a8cf

lvhan028: It'll be OK if the conflicts are resolved.

lvhan028 (approved):
LGTM

PR #1612, opened 2023-01-05
[Fix] Fix aligned allocations on Android

lzhangzz:
Use `posix_memalign` on Android platform, and set minimum alignment to `sizeof(void*)`

- `std::aligned_alloc` is not available in all versions of NDK toolchains.
- Allocation APIs may fail if `alignment` is too small.
- `new` and `malloc` are aligned to `sizeof(void*)` anyway.

https://en.cppreference.com/w/cpp/memory/c/aligned_alloc

### DIFF
--- a/csrc/mmdeploy/device/cpu/cpu_device.cpp
+++ b/csrc/mmdeploy/device/cpu/cpu_device.cpp
@@ -12,12 +12,13 @@ class CpuHostMemory : public NonCopyable {
   CpuHostMemory() : size_(), data_(), owned_data_{false} {}
   Result<void> Init(size_t size, size_t alignment) {
     alignment = std::max(alignment, sizeof(void*));
+    auto space = (size + alignment - 1) / alignment * alignment;
 #ifdef _MSC_VER
-    data_ = _aligned_malloc(size, alignment);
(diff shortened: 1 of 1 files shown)


csrc/mmdeploy/device/cpu/cpu_device.cpp
```diff
@@ -11,11 +11,13 @@ class CpuHostMemory : public NonCopyable {
...
-    size_t space = (size + alignment - 1) / alignment * alignment;
+    alignment = std::max(alignment, sizeof(void*));
 #ifdef _MSC_VER
-    data_ = _aligned_malloc(space, alignment);
+    data_ = _aligned_malloc(size, alignment);
+#elif defined(ANDROID)
+    posix_memalign(&data_, alignment, size);
 #else
-    data_ = std::aligned_alloc(alignment, space);
+    data_ = std::aligned_alloc(alignment, size);
```
grimoire, line 20: `aligned_alloc` require `size - number of bytes to allocate. An integral multiple of alignment.`. Is it safe to send the size without padding?

hanrui1sensetime (approved):
LGTM.

grimoire (approved):
LGTM

PR #1619, opened 2023-01-06
[Improvement] Better unit test.

grimoire:

## Motivation

The unit tests might not release resources if the test failed. Create fixtures for the resources.

## Modification

- Fix rewriter staticmethod
- Better backendswitchwrapper
- resources fixture.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
+++ b/csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
@@ -202,13 +202,13 @@ nvinfer1::IPluginV2 *TRTRoIAlignCreator::createPlugin(
     if (field_name.compare("mode") == 0) {
       int data_size = fc->fields[i].length;
       const char *data_start = static_cast<const char *>(fc->fields[i].data);
-      std::string poolModeStr(data_start);
-      if (poolModeStr == "avg") {
+      std::string pool_mode_str(data_start);
+      if (pool_mode_str == "avg") {
         poolMode = 1;
-      } else if (poolModeStr == "max") {
+      } else if (pool_mode_str == "max") {
         poolMode = 0;
       } else {
-        std::cout << "Unknown pool mode \"" << poolModeStr << "\"." << std::endl;
+        std::cout << "Unknown pool mode \"" << pool_mode_str << "\"." << std::endl;
       }
       ASSERT(poolMode >= 0);
     }
--- a/csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
+++ b/csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
@@ -202,7 +202,7 @@ nvinfer1::IPluginV2 *TRTRoIAlignCreator::createPlugin(
     if (field_name.compare("mode") == 0) {
       int data_size = fc->fields[i].length;
       const char *data_start = static_cast<const char *>(fc->fields[i].data);
-      std::string poolModeStr(data_start, data_size);
+      std::string poolModeStr(data_start);
       if (poolModeStr == "avg") {
         poolMode = 1;
       } else if (poolModeStr == "max") {
--- a/mmdeploy/core/rewriters/function_rewriter.py
+++ b/mmdeploy/core/rewriters/function_rewriter.py
@@ -1,4 +1,5 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+import inspect
 from typing import (Any, Callable, Dict, List, MutableSequence, Optional,
                     Tuple, Union)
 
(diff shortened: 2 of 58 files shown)


csrc/mmdeploy/backend_ops/tensorrt/roi_align/trt_roi_align.cpp
```diff
@@ -202,7 +202,7 @@ nvinfer1::IPluginV2 *TRTRoIAlignCreator::createPlugin(
     if (field_name.compare("mode") == 0) {
       int data_size = fc->fields[i].length;
       const char *data_start = static_cast<const char *>(fc->fields[i].data);
-      std::string poolModeStr(data_start, data_size);
+      std::string poolModeStr(data_start);
```
lvhan028, line 205: may use snake case

mmdeploy/core/rewriters/function_rewriter.py
```diff
@@ -72,7 +73,16 @@ def _set_func(origin_func_path: str,
...
+
+    is_static_method = False
+    if method_class:
+        origin_type = inspect.getattr_static(module_or_class, split_path[-1])
+        is_static_method = isinstance(origin_type, staticmethod)
+
+    if is_static_method:
+        exec(f'{origin_func_path} = staticmethod(rewrite_func)')
+    else:
+        exec(f'{origin_func_path} = rewrite_func')
```
AllentDan, line 85: why do we make difference between static methods and other methods? Removing `self` for beginners might be confusing compared with original way of rewriting.

grimoire: 1. static method does not have self 
https://github.com/open-mmlab/mmdetection3d/blob/47285b3f1e9dba358e98fcd12e523cfd0769c876/mmdet3d/core/bbox/coders/fcos3d_bbox_coder.py#L103
2. static method need special treatment, or it won't recover after context exit.

AllentDan: I see.

tests/test_codebase/test_mmcls/test_classification.py
```diff
@@ -73,58 +81,57 @@ def test_init_pytorch_model(from_mmrazor: Any):
...
-def test_visualize(backend_model):
-    input_dict, _ = task_processor.create_input(img, input_shape=img_shape)
+def test_visualize(task_processor, backend_model, tmp_path, img, model_inputs):
+    input_dict, _ = model_inputs
     results = task_processor.run_inference(backend_model, input_dict)
-    with TemporaryDirectory() as dir:
-        filename = dir + 'tmp.jpg'
-        task_processor.visualize(backend_model, img, results[0], filename, '')
-        assert os.path.exists(filename)
+    filename = str(tmp_path / 'tmp.jpg')
```
RunningLeon, line 118: is it OK to concat string in this way? Better use `os.path` to join directories.

tests/test_codebase/test_mmcls/test_classification_model.py
```diff
@@ -1,123 +1,110 @@
...
         imgs = torch.rand(2, 3, IMAGE_SIZE, IMAGE_SIZE)
-        results = self.end2end_model.forward_test(imgs)
+        results = end2end_model.forward_test(imgs)
         assert isinstance(results[0], np.ndarray)
 
-    def test_show_result(self):
+    def test_show_result(self, end2end_model, tmp_path):
         input_img = np.zeros([IMAGE_SIZE, IMAGE_SIZE, 3])
-        img_path = NamedTemporaryFile(suffix='.jpg').name
+        img_path = str(tmp_path / 'tmp.jpg')
```
RunningLeon, line 58: here also

tests/test_codebase/test_mmdet3d/test_mmdet3d_models.py
```diff
@@ -113,30 +117,34 @@ def test_pointpillars_scatter(backend_type: Backend):
...
 
 
-def get_centerpoint_head():
-    from mmdet3d.models import builder
-    model_cfg.centerpoint_model.pts_bbox_head.test_cfg = model_cfg.\
-        centerpoint_model.test_cfg
-    head = builder.build_head(model_cfg.centerpoint_model.pts_bbox_head)
-    head.requires_grad_(False)
-    return head
+# @pytest.fixture
```
RunningLeon, line 130: should we remove this ut?

tests/test_codebase/test_mmocr/test_mmocr_models.py
```diff
@@ -506,7 +503,7 @@ def test_sar_model(backend: Backend, decoder_type):
     pytorch_model.cfg = sar_cfg
     patched_model = patch_model(
         pytorch_model, cfg=deploy_cfg, backend=backend.value)
-    onnx_file_path = tempfile.NamedTemporaryFile(suffix='.onnx').name
+    onnx_file_path = str(tmp_path / 'tmp.onnx')
```
RunningLeon, line 506: does this work for windows?

tests/test_codebase/test_mmcls/test_classification.py
grimoire: [tmp_path](https://docs.pytest.org/en/7.2.x/reference/reference.html#tmp-path) is a builtin fixture in pytest which return a [pathlib.Path](https://docs.python.org/3/library/pathlib.html#pathlib.Path) object.

tests/test_codebase/test_mmocr/test_mmocr_models.py
grimoire: [tmp_path](https://docs.pytest.org/en/7.2.x/reference/reference.html#tmp-path) is a builtin fixture in pytest which return a [pathlib.Path](https://docs.python.org/3/library/pathlib.html#pathlib.Path) object.

AllentDan (approved):

RunningLeon (approved):
LGTM

Issue #1153, opened 2022-10-03
How to convert a segmentation model to tensorrt engine that can output probability map ?


yuanyuangoo:
Hi, Now the tensorrt engine can output result that has been passed through argmax function, how to also get output of the seg_logit before that?

RunningLeon:
@yuanyuangoo Hi, a simple way is to return the seg_logit in the rewritings function, just like [here](https://github.com/open-mmlab/mmdeploy/blob/b2cb5722f732c4789a11e282d13b00b6e25e2b0d/mmdeploy/codebase/mmseg/models/segmentors/encoder_decoder.py#L35).

yuanyuangoo:
Hi, Thanks. It works to get seg_logit. Then How can I modify csrc/mmdeploy/codebase/mmseg/segment.cpp to accommodate the new output size? It gives me error like "[2022-10-10 08:50:36.869] [mmdeploy] [error] [segment.cpp:35] unsupported `output` tensor, shape: [1, 8, 512, 1024]"

RunningLeon:
@yuanyuangoo Hi, if you want to support this in sdk, you have to add  another `post_process` with `resize and argmax`.

AllentDan:
#1379 might be what you need. @yuanyuangoo

Issue #887, opened 2022-08-10
TensorRT segmentation probabilities scores


user41pp:
**Describe the feature**

MMSegmentation TensorRT deployment: output float probabilities/scores instead of int class labels.

In segmentor.h we have
`int* mask;    ///< segmentation mask of the input image, in which mask[i * width + j] indicates the label id of pixel at (i, j)`

Request: change mask to be a 3d matrix of probabilities of shape (width, height, num_classes) with float probabilities for all classes.

This would require modifications in the export pipeline as well.

**Motivation**

- Variable thresholds for different classes at runtime
- Decision making based on all scores, not just the top one, if available

**Related resources**

https://github.com/open-mmlab/mmdeploy/blob/master/csrc/mmdeploy/apis/c/mmdeploy/segmentor.h

RunningLeon:
@user41pp Hi, exporting to onnx with score map(without argmax) is OK and easy to implement, but setting thresholds for each class at runtime is a little confused. Maybe you could provide sample code or PR to show your thoughts.

user41pp:
> @user41pp Hi, exporting to onnx with score map(without argmax) is OK and easy to implement, but setting thresholds for each class at runtime is a little confused. Maybe you could provide sample code or PR to show your thoughts.

Hi, sorry for the late reply. In the simplest form, I am asking for exactly that - adding the option (or making it the default) to remove the argmax layer and instead directly output the softmax values.

Here is a screenshot of the last layers of a UNET model (configs and commands listed further below) with that change:

![unet_fcn_output_new](https://user-images.githubusercontent.com/4224596/184653449-a2b0b4de-384c-47e3-8cba-f6683436259b.jpg)


This would solve our problem already and if that's easy, then this is the way to go. The "dynamic threshold at runtime" part does not have to be done by mmdeploy.

However, outputting the softmax layer would increase the output tensor by the number of classes the model has (e.g. current output shape is 1024 x 512, and if we would output the softmax values, output shape would be 1024 x 512 x num_classes). Because of the potentially huge amount of classes maybe it makes sense to not output all the scores but instead only the topk scores and correspondig class indices. For the most common case k=1 we would output the top softmax score along with its class index. The returned tensor then would be a tuple containing (argmax, softmax[argmax]).

MMdeploy config: 
https://github.com/open-mmlab/mmdeploy/blob/master/configs/mmseg/segmentation_tensorrt-fp16_dynamic-512x1024-2048x2048.py

[...]

PR #1625, opened 2023-01-09
[Feature] Support feature map output for mmsegmentation

irexyc:
## Motivation

Add feature map output for mmsegmentation (sdk)

## Modification

- [x] Add score filed for mmdeploy_segmentation_t
- [x] update c/cpp api && demo
- [x] update csharp api && demo
- [x] update python api && demo

## Related issues & pr
https://github.com/open-mmlab/mmdeploy/pull/1379
https://github.com/open-mmlab/mmdeploy/issues/887
https://github.com/open-mmlab/mmdeploy/issues/1153


csrc/mmdeploy/codebase/mmseg/segment.cpp
```diff
@@ -41,40 +45,60 @@ class ResizeMask : public MMSegmentation {
...
+        cv::extractChannel(mask_mat, mask_mat, little_endian_ ? 0 : mask_mat.channels() - 1);
+      }
+      if (mask_mat.type() != CV_32S) {
+        mask_mat.convertTo(mask_mat, CV_32S);
+      }
+      resized_mask = cpu::Resize(mask_mat, input_height, input_width, "nearest");
+      tensor_mask = cpu::CVMat2Tensor(resized_mask);
+    } else {
+      // score
+      resized_score = cpu::Resize(mask_mat, input_height, input_width, "bilinear");
```
lzhangzz, line 89: `cpu::Resize` does not support `C>4`

demo/csrc/c/image_segmentation.cpp
```diff
@@ -60,7 +60,19 @@ int main(int argc, char* argv[]) {
...
   int pos = 0;
   for (auto iter = color_mask.begin<cv::Vec3b>(); iter != color_mask.end<cv::Vec3b>(); ++iter) {
-    *iter = palette[result->mask[pos++]];
+    // output mask
+    if (result->mask) {
+      *iter = palette[result->mask[pos++]];
+    }
+    // output score
+    if (result->score) {
+      std::vector<std::pair<float, int>> score(result->classes);
```
lzhangzz, line 69: This is going to be slow.
```c++
std::vector<int> idxs(result->classes);
std::iota(idxs.begin(), idsx.end(), 0);
auto k = std::max_element(idxs.begin(), idex.end(), [&](int i, int j){ 
    return result->score[pos+i] < result->score[pos+j]; 
}) - idxs.begin();
*iter = palette[k];
```
and move the definition of `idxs` out of the loop.

csrc/mmdeploy/codebase/mmseg/segment.cpp
```diff
@@ -41,40 +45,60 @@ class ResizeMask : public MMSegmentation {
...
+    auto channel = (int)mask.shape(1);
     auto height = (int)mask.shape(2);
     auto width = (int)mask.shape(3);
     auto input_height = preprocess_result["img_metas"]["ori_shape"][1].get<int>();
     auto input_width = preprocess_result["img_metas"]["ori_shape"][2].get<int>();
+    if (!with_argmax_) {
+      // (C, H, W) -> (H, W, C)
+      ::mmdeploy::operation::Context ctx(device_, stream_);
+      std::vector<int> axes = {0, 2, 3, 1};
+      permute_.Apply(mask, mask, axes);
```
lzhangzz, line 62: IMO, CHW format score maps are more intuitive for the users.

csrc/mmdeploy/codebase/mmseg/segment.cpp
irexyc: It seems support channel <= 512

csrc/mmdeploy/codebase/mmseg/segment.cpp
```diff
@@ -41,40 +45,60 @@ class ResizeMask : public MMSegmentation {
...
+      if (mask_mat.channels() > 1) {
+        cv::extractChannel(mask_mat, mask_mat, little_endian_ ? 0 : mask_mat.channels() - 1);
+      }
+      if (mask_mat.type() != CV_32S) {
+        mask_mat.convertTo(mask_mat, CV_32S);
+      }
+      resized_mask = cpu::Resize(mask_mat, input_height, input_width, "nearest");
+      tensor_mask = cpu::CVMat2Tensor(resized_mask);
+    } else {
+      // score
```
munhou, line 88: can you take a look [1648](https://github.com/open-mmlab/mmdeploy/issues/1648)

I google this error, someone said that cv::resize does not support int32, so I changed the code of 0.12.0 to the following, it seems to work well, but I haven't tried the code of this branch, because I think it can't solve my problem
```
    if (mask_mat.type() != CV_32F) {
      mask_mat.convertTo(mask_mat, CV_32F);
    }

    cv::Mat resized_mask = cpu::Resize(mask_mat, input_height, input_width, "nearest");
```
Another thing is strange, the resize of CV_32S on 0.10.0 is correct, I compared the input before resize of the two versions, they are the same, 
```
[2023-01-13 17:49:00.016] [mmdeploy] [info] [segment.cpp:33] tensor.name: output, tensor.shape: [1, 1, 512, 512], tensor.data_type: INT32
```

but the version of ppl.cv used by 0.10.0 and 0.12.0 is also the same, I can't find the original cause of my problem.
I will keep trying to find the reason, if you have any suggestions, or this branch has been able to solve my problem，please let me know

munhou: Thank you for your patience，I saw that it uses opencv, but I only noticed the version of opencv, and it was installed through ppl.cv, I didn't notice the call parameter of resize，Thanks again，With your guidance, I have solved my problem

hanrui1sensetime (approved):
LGTM.

csrc/mmdeploy/codebase/mmseg/segment.cpp
lzhangzz: And this also contradicts with the fact that the model's result is in NCHW format.

irexyc: CHW format may be inconvenient to use opencv resize if the user want to resize the output feature map.

csrc/mmdeploy/apis/c/mmdeploy/segmentor.h
```diff
@@ -17,11 +17,14 @@ extern "C" {
...
 typedef struct mmdeploy_segmentation_t {
-  int height;   ///< height of \p mask that equals to the input image's height
-  int width;    ///< width of \p mask that equals to the input image's width
-  int classes;  ///< the number of labels in \p mask
-  int* mask;    ///< segmentation mask of the input image, in which mask[i * width + j] indicates
-                ///< the label id of pixel at (i, j)
+  int height;    ///< height of \p mask that equals to the input image's height
+  int width;     ///< width of \p mask that equals to the input image's width
+  int classes;   ///< the number of labels in \p mask
+  int* mask;     ///< segmentation mask of the input image, in which mask[i * width + j] indicates
```
lvhan028, line 23: Please clarify if `mask` or `score` might be null

csrc/mmdeploy/apis/c/mmdeploy/segmentor.h
```diff
@@ -17,11 +17,14 @@ extern "C" {
...
+                 ///< the label id of pixel at (i, j)
+  float* score;  ///< segmentation score map of the input image, in which
+                 ///< score[height * width * k + i * width + j] indicates the score
```
lvhan028, line 26: what does `k` mean in `[height * width * k + i * width + j]`

csrc/mmdeploy/codebase/mmaction/CMakeLists.txt
```diff
@@ -4,8 +4,7 @@ project(mmdeploy_mmaction)
 
 file(GLOB SRCS ${CMAKE_CURRENT_SOURCE_DIR} "*.cpp")
 mmdeploy_add_module(${PROJECT_NAME} "${SRCS}")
-add_subdirectory(cpu)
```
lvhan028, line 7: Why remove them in this PR?

demo/python/image_segmentation.py
```diff
@@ -35,6 +35,8 @@ def main():
     segmentor = Segmentor(
         model_path=args.model_path, device_name=args.device_name, device_id=0)
     seg = segmentor(img)
+    if seg.dtype == np.float32:
+        seg = np.argmax(seg, axis=2)
```
lvhan028, line 39: `axis=2`?

csrc/mmdeploy/operation/cpu/permute.cpp
```diff
@@ -0,0 +1,83 @@
...
+namespace mmdeploy::operation::cpu {
+
+class PermuteImpl : public Permute {
+ public:
+  explicit PermuteImpl() {}
+
+  Result<void> apply(const Tensor& src, Tensor& dst, const std::vector<int>& axes) override {
+    int ndim = src.shape().size();
+    if (ndim != axes.size()) {
+      MMDEPLOY_ERROR("The size of axes should be equal of src, {} vs {}", axes.size(), ndim);
```
lvhan028, line 15: 'be equal to'

csrc/mmdeploy/operation/cpu/permute.cpp
```diff
@@ -0,0 +1,83 @@
...
+    int ndim = src.shape().size();
+    std::vector<int> coord(ndim, 0);
+    auto dst_data = dst.data<T>();
+    auto src_data = src.data<T>();
+
+    int i;
+    do {
+      dst_data[0] = src_data[0];
+      for (i = ndim - 1; i >= 0; i--) {
+        if (++coord[i] == shape[i]) {
```
lvhan028, line 66: what if there are duplicated numbers in `shape`?

csrc/mmdeploy/codebase/mmseg/segment.cpp
```diff
@@ -20,6 +22,8 @@ class ResizeMask : public MMSegmentation {
       classes_ = cfg["params"]["num_classes"].get<int>();
       with_argmax_ = cfg["params"].value("with_argmax", true);
       little_endian_ = IsLittleEndian();
+      ::mmdeploy::operation::Context ctx(Device("cpu"), stream_);
```
lvhan028, line 25: What's `ctx` used for?

csrc/mmdeploy/codebase/mmseg/segment.cpp
```diff
@@ -41,40 +45,62 @@ class ResizeMask : public MMSegmentation {
...
     auto height = (int)mask.shape(2);
     auto width = (int)mask.shape(3);
     auto input_height = preprocess_result["img_metas"]["ori_shape"][1].get<int>();
     auto input_width = preprocess_result["img_metas"]["ori_shape"][2].get<int>();
     Device host{"cpu"};
     OUTCOME_TRY(auto host_tensor, MakeAvailableOnDevice(mask, host, stream_));
-    OUTCOME_TRY(stream_.Wait());
+    if (!with_argmax_) {
+      // (C, H, W) -> (H, W, C)
+      ::mmdeploy::operation::Context ctx(host, stream_);
```
lvhan028, line 62: What's `ctx` used for?

csrc/mmdeploy/codebase/mmaction/format_shape.cpp
```diff
@@ -127,6 +130,14 @@ Result<void> FormatShape::Apply(Value& data) {
 
 MMDEPLOY_REGISTER_TRANSFORM(FormatShape);
 
-MMDEPLOY_DEFINE_REGISTRY(FormatShapeOp);
+MMDEPLOY_DEFINE_REGISTRY(FormatShapeImpl);
```
lvhan028, line 133: Why not merge `FormatShapeImpl` and `FormatShape`?
BTW, I think it's better to open another PR to refactor mmaction's format shape.
It can reduce review workload

csrc/mmdeploy/operation/cpu/permute.cpp
irexyc: Do you mean duplicated  axes?

lvhan028:
How to test it?

lvhan028:
May add TC for `permute`

lzhangzz (approved):

lvhan028 (approved):

PR #1626, opened 2023-01-09
[Fix] Fix example standalone build for msvc

irexyc:
## Motivation

Fix example standalone build for msvc

### DIFF
--- a/csrc/mmdeploy/core/CMakeLists.txt
+++ b/csrc/mmdeploy/core/CMakeLists.txt
@@ -54,6 +54,11 @@ target_include_directories(${PROJECT_NAME}
         $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/third_party/json>
         )
 
+if (MSVC)
+    target_compile_options(${PROJECT_NAME} PUBLIC
+            $<$<COMPILE_LANGUAGE:CXX>:/Zc:preprocessor>)
+endif ()
+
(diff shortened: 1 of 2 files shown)


csrc/mmdeploy/core/CMakeLists.txt
```diff
@@ -54,6 +54,11 @@ target_include_directories(${PROJECT_NAME}
         $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/third_party/json>
         )
 
+if (MSVC)
```
lvhan028, line 57: Looks `Zc:preprocessor` is moved from root cmakelists here
What's the difference?

csrc/mmdeploy/core/CMakeLists.txt
```diff
@@ -54,6 +54,11 @@ target_include_directories(${PROJECT_NAME}
...
+    target_compile_options(${PROJECT_NAME} PUBLIC
+            $<$<COMPILE_LANGUAGE:CXX>:/Zc:preprocessor>)
+endif ()
```
lzhangzz, line 60: We'd better also move `add_compile_options($<$<COMPILE_LANGUAGE:CXX>:/Zc:__cplusplus>)` here since `__cplusplus` is may not be correctly defined on legacy MSVC without the flag.

lzhangzz (approved):

lvhan028 (approved):

PR #1636, opened 2023-01-11
[Fix] fix rknn quantization pipeline

AllentDan:


lvhan028 (approved):

tools/deploy.py
```diff
@@ -218,7 +218,8 @@ def main():
         dataset_file = tempfile.NamedTemporaryFile(suffix='.txt').name
         with open(dataset_file, 'w') as f:
             f.writelines([osp.abspath(args.img)])
-        quantization_cfg.setdefault('dataset', dataset_file)
+        if quantization_cfg['dataset'] is None:
```
grimoire, line 221: What if the key 'dataset' does not exist?

grimoire (approved):

PR #1704, opened 2023-02-02
[Fix] Fix dbnet regression test

AllentDan:
Directly set pytorch metric when it's empty for regression_test.py.

### DIFF
--- a/tools/regression_test.py
+++ b/tools/regression_test.py
@@ -316,7 +316,7 @@ def get_pytorch_result(model_name: str, meta_info: dict, checkpoint_path: Path,
                     pytorch_metric.update(_metrics[ds])
                 if ds == dataset:
                     pytorch_metric.update(_metrics)
-        else:
+        if pytorch_metric == dict():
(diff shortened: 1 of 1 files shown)


grimoire (approved):

AllentDan:
@RunningLeon Would this change the behavior of MMDet? I heard that the MMDet coco dataset stuff is sensitive to multiple datasets settings.

RunningLeon:
> @RunningLeon Would this change the behavior of MMDet? I heard that the MMDet coco dataset stuff is sensitive to multiple datasets settings.

Not sure, maybe you could start a fast test with jenkins.

AllentDan:
> > @RunningLeon Would this change the behavior of MMDet? I heard that the MMDet coco dataset stuff is sensitive to multiple datasets settings.
> 
> Not sure, maybe you could start a fast test with jenkins.

Tested OK with 1660 card.

RunningLeon (approved):
LGTM

PR #1705, opened 2023-02-02
[fix] fix build_backend_model() error

AllentDan:


AllentDan:
@RunningLeon Would this change the behavior of MMDet? I heard that the MMDet coco dataset stuff is sensitive to multiple datasets settings.

grimoire (approved):

PR #1708, opened 2023-02-03
[Fix] fix torch allocator resouce releasing

AllentDan:
Reproduce codes in master branch:
```python
from mmdeploy.backend.tensorrt.wrapper import TRTWrapper
trt_model = TRTWrapper(
    '/path_to/cls/resnet/end2end.engine',
    ['output'])
input = torch.rand(1, 3, 224, 224).cuda()
output_trt = trt_model.forward({'input': input})
print(output_trt)
```
Error might triggered:
```
/mmdeploy/utils/logging.py(28): get_logger
Exception caught in deallocate(): TypeError: 'NoneType' object is not callable
```
or
```
codes/mmdeploy/mmdeploy/backend/tensorrt/torch_allocator.py(61): deallocate

[ERROR] Exception caught in deallocate(): AttributeError: 'NoneType' object has no attribute 'caching_allocator_delete'
```

### DIFF
--- a/mmdeploy/backend/tensorrt/torch_allocator.py
+++ b/mmdeploy/backend/tensorrt/torch_allocator.py
@@ -13,6 +13,7 @@ class TorchAllocator(trt.IGpuAllocator):
 
         self.device_id = device_id
         self.mems = set()
+        self.caching_delete = torch._C._cuda_cudaCachingAllocator_raw_delete
 
     def __del__(self):
         """destructor."""
@@ -56,7 +57,6 @@ class TorchAllocator(trt.IGpuAllocator):
         if memory not in self.mems:
             return False
 
-        if hasattr(torch.cuda, 'caching_allocator_delete'):
-            torch.cuda.caching_allocator_delete(memory)
+        self.caching_delete(memory)
         self.mems.discard(memory)
         return True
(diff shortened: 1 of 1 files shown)


mmdeploy/backend/tensorrt/torch_allocator.py
```diff
@@ -53,11 +53,10 @@ def deallocate(self: trt.IGpuAllocator, memory: int) -> bool:
         Returns:
             bool: deallocate success.
         """
-        logger = get_root_logger()
```
lvhan028, line 56: Can we still keep them if mmengine is installed?

AllentDan: Yes. We can also modify the codes of [get_logger()](https://github.com/open-mmlab/mmdeploy/blob/master/mmdeploy/utils/logging.py).

grimoire:
Unable to reproduce the error.
Why would `logging` and `torch.cuda` be None?

AllentDan:
> Unable to reproduce the error. Why would `logging` and `torch.cuda` be None?

Don't know either. Just happened at the end of the program. Would it be possible that torch.cuda and logging are released at that time.

grimoire:
More investigations are required.

grimoire (approved):

grimoire:
Could report to TensorRT team.

lvhan028 (approved):

PR #1548, opened 2022-12-16
fix(README): badge error and add github action status

tpoisonooo:
Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily receiving feedbacks. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation

fix readme error

## Modification

* update URL
* github action add read permission

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
3. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #1713, opened 2023-02-06
fix stale workflow

RunningLeon:
## Motivation
Stale workflow should have `write` permission to do job such close issues per [here](https://github.com/actions/stale#recommended-permissions).

## Modification

Reset to default permission, which changed in PR https://github.com/open-mmlab/mmdeploy/pull/1548

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -5,9 +5,6 @@ on:
     # check issue and pull request once at 01:30 a.m. every day
     - cron: '30 1 * * *'
 
-permissions:
-  contents: read
-
 jobs:
   stale:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


lvhan028 (approved):

.github/workflows/stale.yml
```diff
@@ -5,9 +5,6 @@ on:
     # check issue and pull request once at 01:30 a.m. every day
     - cron: '30 1 * * *'
 
-permissions:
-  contents: read
```
lvhan028, line 9: Can we use 'write' explicitly to avoid changes by mistake?

RunningLeon: yes

tpoisonooo (approved):

PR #1718, opened 2023-02-07
[Feature] Sync csharp apis with newly added c apis && demo

irexyc:
- Component 
  - [x] `Context`
  - [x] `Device`
  - [x] `Model`
  - [x] `Scheduler`
  - [x] `Profiler`
- API
  - [x] PoseTracker`
  - [x] RotatedDetector
- Demo
  - [x] Pose tracking
  - [x] Rotated detection.


demo/csharp/pose_tracking/Program.cs
```diff
@@ -0,0 +1,199 @@
...
+                mat.Channel = cvMat.Dims;
+                mat.Format = PixelFormat.BGR;
+                mat.Type = DataType.Int8;
+                mat.Device = null;
+            }
+        }
+
+        static void PrintHelperMessage()
+        {
+            string message = "usage:\n pose_tracking device det_model pose_model video";
```
lvhan028, line 128: May change `pose_tracking` to `pose_tracker` for consistency

demo/csharp/pose_tracking/Program.cs
```diff
@@ -0,0 +1,199 @@
...
+                if (frame.Empty())
+                {
+                    break;
+                }
+                CvMatToMat(frame, out var mat);
+                // process
+                PoseTrackerOutput result = tracker.Apply(state, mat);
+
+                // visualize
+                if (!Visualize(frame, result, 1280, frame_id++, true))
```
lvhan028, line 189: I don't think a hardcode `1280` is a good idea.
If a user-specified size is not an argument for this demo, the original mat size sounds more reasonable

lvhan028 (approved):

lzhangzz (approved):

PR #1739, opened 2023-02-09
[Enhance] support TensorRT engine for onnxruntime

yhna940:

## Motivation

With the TensorRT execution provider, the ONNX Runtime delivers better inferencing performance on the same hardware compared to generic GPU acceleration. The TensorRT execution provider in the ONNX Runtime makes use of NVIDIA’s [TensorRT](https://developer.nvidia.com/tensorrt) Deep Learning inferencing engine to accelerate ONNX model in their family of GPUs. 

https://onnxruntime.ai/docs/execution-providers/TensorRT-ExecutionProvider.html

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdeploy/backend/onnxruntime/wrapper.py
+++ b/mmdeploy/backend/onnxruntime/wrapper.py
@@ -50,16 +50,16 @@ class ORTWrapper(BaseWrapper):
             logger.warning('The library of onnxruntime custom ops does '
                            f'not exist: {ort_custom_op_path}')
         device_id = parse_device_id(device)
-        providers = ['CPUExecutionProvider']
+        providers = []
         if 'cuda' in device:
-            providers.insert(0, ('CUDAExecutionProvider', {
-                'device_id': device_id
-            }))
             if 'TensorrtExecutionProvider' in ort.get_available_providers():
-                providers.insert(0, ('TensorrtExecutionProvider', {
+                providers.append(('TensorrtExecutionProvider', {
                     'device_id': device_id
                 }))
-
+            providers.append(('CUDAExecutionProvider', {
+                'device_id': device_id
+            }))
+        providers.append('CPUExecutionProvider')
         sess = ort.InferenceSession(
             onnx_file, session_options, providers=providers)
         if output_names is None:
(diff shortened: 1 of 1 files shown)


mmdeploy/backend/onnxruntime/wrapper.py
```diff
@@ -36,7 +38,8 @@ class ORTWrapper(BaseWrapper):
     def __init__(self,
                  onnx_file: str,
                  device: str,
-                 output_names: Optional[Sequence[str]] = None):
+                 output_names: Optional[Sequence[str]] = None,
+                 enable_trt: bool = False):
```
grimoire, line 42: Is it possible to check if TensorRT provider is available in the current environment?
If it is, we can check it inside init instead of add a flag.

yhna940:
```
>>> available_providers = onnxruntime.get_available_providers()
>>> available_providers
['TensorrtExecutionProvider', 'CUDAExecutionProvider', 'CPUExecutionProvider']
```

`onnxruntime` has a `get_available_providers` function that tells you which providers are available in the current environment. Changed to use TensorRT executor if available. thank you :)

mmdeploy/backend/onnxruntime/wrapper.py
```diff
@@ -50,9 +50,17 @@ def __init__(self,
...
                            f'not exist: {ort_custom_op_path}')
         device_id = parse_device_id(device)
-        providers = ['CPUExecutionProvider'] \
-            if device == 'cpu' else \
-            [('CUDAExecutionProvider', {'device_id': device_id})]
+        providers = []
+        if device == 'cpu':
+            providers.append('CPUExecutionProvider')
+        else:
+            providers.append(('CUDAExecutionProvider', {
```
grimoire, line 57: As far as I know, cpu provider can co-exist with cuda provider, ONNXRuntime will fall back to cpu if the cuda implementation of the op is not provided.
Cpu provider can be placed in the providers list even when we use gpu device. (I am not sure if the order of the provider matters)

yhna940: According to the official documentation, the order of providers is the same as preference. Therefore, when the device is cuda, tensorrt and cuda executor are included, and the cpu executor is added at the end as the default value. (https://onnxruntime.ai/docs/execution-providers/)

grimoire (approved):
LGTM

lzhangzz (approved):

PR #1740, opened 2023-02-09
[Enhancement] Support cmake configure when system exists multiple cuda versions.

irexyc:
## Motivation

On windows, to support cmake use cuda as a language, one need to install cuda msbuild extensions that cuda installer provides.

When configure mmdeploy with cuda, if not specify cuda toolkit, cmake will use the highest cuda extensions found in `$(VCTargetsPath)\BuildCustomizations`. And it also need environment variable `CUDA_PATH_V{X}_{Y}` to be set.

When using `find_package(CUDA)`, it will first search `CUDA_PATH` and then the dir of `nvcc.exe` if `nvcc.exe` can be found. The search path should be compatible with used extensions.

## Usage.
- If no cuda msbuild extensions installed to vs folder
  ```bash
  cmake  -A x64 -T v142,cuda="/path/to/cuda" ..
  ```
- If multiple cuda msbuild extensions installed to vs folder
  ```bash
  cmake  -A x64 -T v142,cuda="/path/to/cuda" ..
  OR
  cmake -A x64 -T v142,cuda=x.y
  ```

### DIFF
--- a/cmake/cuda.cmake
+++ b/cmake/cuda.cmake
@@ -31,7 +31,7 @@ if (MSVC)
         string(REGEX REPLACE ".*v[0-9]+\\.([0-9]+).*" "\\1" _MINOR $ENV{CUDA_PATH})
         if (NOT (${CMAKE_VS_PLATFORM_TOOLSET_CUDA} STREQUAL "${_MAJOR}.${_MINOR}"))
             message(FATAL_ERROR "Auto detected cuda version ${CMAKE_VS_PLATFORM_TOOLSET_CUDA}"
-                " is mismatch with ENV{CUDA_PATH} $ENV{CUDA_PATH}. Plese modify CUDA_PATH"
+                " is mismatch with ENV{CUDA_PATH} $ENV{CUDA_PATH}. Please modify CUDA_PATH"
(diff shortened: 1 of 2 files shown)


cmake/cuda.cmake
```diff
@@ -11,6 +11,37 @@ if (MSVC OR (NOT DEFINED CMAKE_CUDA_RUNTIME_LIBRARY))
...
 endif ()
 
+if (MSVC)
+    # no plugin in BuildCustomizations and no specify cuda toolset
+    if (NOT CMAKE_VS_PLATFORM_TOOLSET_CUDA)
+        message(FATAL_ERROR "Please install CUDA MSBuildExtensions")
+    endif ()
+
+    if (CMAKE_VS_PLATFORM_TOOLSET_CUDA_CUSTOM_DIR)
+        # find_package(CUDA) required ENV{CUDA_PATH}
```
lvhan028, line 21: may remove this comment

RunningLeon (approved):
LGTM

lvhan028 (approved):

lzhangzz (approved):

PR #1755, opened 2023-02-14
Fix trtlogger instead of mm logger

triple-Mu:
TensorRT's logger behaves differently from mmdeploy's logger.
``` shell
>>> import tensorrt as trt
>>> log_level = trt.Logger.ERROR
>>> logger = trt.Logger(log_level)
>>> logger.warning('This is a warning')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'tensorrt.tensorrt.Logger' object has no attribute 'warning'
>>> logger.log(log_level, 'This is a warning')
[02/14/2023-11:26:54] [TRT] [E] This is a warning
```
We should use `logger.log` method for printing infomation.
Maybe this pr can also applies to the dev branch?

### DIFF
--- a/mmdeploy/backend/tensorrt/utils.py
+++ b/mmdeploy/backend/tensorrt/utils.py
@@ -138,7 +138,7 @@ def from_onnx(onnx_model: Union[str, onnx.ModelProto],
         >>>             device_id=0)
         >>>             })
     """
-
+    from functools import partial
     if device_id != 0:
         import os
         old_cuda_device = os.environ.get('CUDA_DEVICE', None)
@@ -152,6 +152,7 @@ def from_onnx(onnx_model: Union[str, onnx.ModelProto],
     load_tensorrt_plugin()
     # create builder and network
     logger = trt.Logger(log_level)
+    trtlogger = partial(logger.log, log_level)
     builder = trt.Builder(logger)
 
     # TODO: use TorchAllocator as builder.gpu_allocator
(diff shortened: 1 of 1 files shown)


mmdeploy/backend/tensorrt/utils.py
```diff
@@ -205,18 +206,18 @@ def from_onnx(onnx_model: Union[str, onnx.ModelProto],
         max_shape = param['max_shape']
         profile.set_shape(input_name, min_shape, opt_shape, max_shape)
     if config.add_optimization_profile(profile) < 0:
-        logger.warning(f'Invalid optimization profile {profile}.')
+        trtlogger(f'Invalid optimization profile {profile}.')
```
AllentDan, line 209: This might be a warning instead of a log according to the level of `log_level`?

triple-Mu: We can not control the log level.
If we set  log_level=trt.Logger.ERROR,
when we set logger.log(trt.Logger.Warning, 'message'), 
it take no effects.

AllentDan: The code line `logger.warning(f'Invalid optimization profile {profile}.')` is just to give a warning to the user. So, give a warning `logger.log(trt.Logger.Warning, 'message')`.

AllentDan: I'm not sure here since if the user use the log level *INFO*, the warning here would be info then. Or, we can use mm logger here to do warning? @RunningLeon what's your opinion?

RunningLeon: > I'm not sure here since if the user uses the log level _INFO_, the warning here would be info then. Or, we can use mm logger here to do warning? @RunningLeon what's your opinion?

Agree, better use logger from mmdeploy.

mmdeploy/backend/tensorrt/utils.py
```diff
@@ -152,6 +152,7 @@ def from_onnx(onnx_model: Union[str, onnx.ModelProto],
     load_tensorrt_plugin()
     # create builder and network
     logger = trt.Logger(log_level)
```
RunningLeon, line 154: we could save this logger as trt_logger only for `trt.onnxParser` and create another logger provided by mmdeploy

AllentDan (approved):
LGTM

mmdeploy/backend/tensorrt/utils.py
```diff
@@ -163,6 +162,9 @@ def from_onnx(onnx_model: Union[str, onnx.ModelProto],
     # parse onnx
     parser = trt.OnnxParser(network, logger)
```
RunningLeon, line 163: We'd better rename the variable name of `logger` to `trt_logger`

triple-Mu: > We'd better rename the variable name of `logger` to `trt_logger`

Fixed

RunningLeon (approved):
LGTM

PR #1767, opened 2023-02-15
[Fix] fix MRCNN visualization, add condition for mask resize

AllentDan:


hanrui1sensetime (approved):
LGTM.

grimoire (approved):
LGTM. It is worth mentioning that solo is dead too. But I think it has nothing todo with this PR.

PR #1776, opened 2023-02-17
[Feature] Dynamically load net module to remove dependencies of mmdeploy.so

irexyc:
## Motivation

Move inference engine library dependencies from `mmdeploy.so` to `net module`

PR #1798, opened 2023-02-22
[Fix] Export `mmdeploy` only in monolithic build

lzhangzz:
base on #1797


lzhangzz:
The current design of this PR will break #1776, which has higher priority.

irexyc (approved):

lvhan028 (approved):

Issue #1796, opened 2023-02-22
Error occurred in cross compilation[Bug] 


Dark1Forest:
### Checklist

- [X] I have searched related issues but cannot get the expected help.
- [X] 2. I have read the [FAQ documentation](https://github.com/open-mmlab/mmdeploy/blob/master/docs/en/faq.md) but cannot get the expected help.
- [X] 3. The bug has not been fixed in the latest version.

### Describe the bug

An error is reported when the following command is executed
`
sh -x tools/scripts/ubuntu_cross_build_aarch64.sh
`

### Reproduction

sh -x tools/scripts/ubuntu_cross_build_aarch64.sh

### Environment

```Shell
python tools/check_env.py
Traceback (most recent call last):
  File "/mnt/workspace/mmdeploy/tools/check_env.py", line 4, in <module>
    from mmcv.utils import get_git_hash
ImportError: cannot import name 'get_git_hash' from 'mmcv.utils' (/mnt/assembly/anaconda3/envs/torch2/lib/python3.9/site-packages/mmcv/utils/__init__.py)
```

### Error traceback

```Shell
+ make -j29
Consolidate compiler generated dependencies of target mmdeploy_ncnn_ops_obj
Consolidate compiler generated dependencies of target spdlog
[  3%] Built target mmdeploy_ncnn_ops_obj
[  7%] Built target spdlog
[  8%] Built target mmdeploy_ncnn_ops
Consolidate compiler generated dependencies of target mmdeploy_core
[  8%] Building CXX object csrc/mmdeploy/core/CMakeFiles/mmdeploy_core.dir/module.cpp.o
[  9%] Building CXX object csrc/mmdeploy/core/CMakeFiles/mmdeploy_core.dir/model.cpp.o
[ 10%] Building CXX object csrc/mmdeploy/core/CMakeFiles/mmdeploy_core.dir/net.cpp.o
[ 11%] Building CXX object csrc/mmdeploy/core/CMakeFiles/mmdeploy_core.dir/graph.cpp.o
In file included from /mnt/workspace/mmdeploy/csrc/mmdeploy/core/net.h:7,
                 from /mnt/workspace/mmdeploy/csrc/mmdeploy/core/net.cpp:3:
[...]

PR #1806, opened 2023-02-23
[Fix] Fix Debian aarch64 cross compiling

lzhangzz:
workaround for #1796


lzhangzz:
@lvhan028 This is a workaround for a bug in gcc-10.2, not a bug of ours.

irexyc (approved):

PR #1851, opened 2023-03-08
[Enhancement] rewrite torch.cat for TensorRT when input is dynamic

AllentDan:


mmdeploy/pytorch/functions/cat.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+@FUNCTION_REWRITER.register_rewriter(func_name='torch.cat', backend='tensorrt')
+def cat__tensorrt(tensors: Sequence[Tensor], *args, **kwargs) -> torch.Tensor:
+    """Rewrite `cat` for TensorRT backend.
+
+    cat in TensorRT does not support bool or uint8 type when input is dynamic.
+    """
+    ctx = FUNCTION_REWRITER.get_context()
+    if get_dynamic_axes(ctx.cfg) is None:
```
grimoire, line 18: warning: there is a potential risk. deploy config is not required for tensorrt.from_onnx.

grimoire (approved):
LGTM

mmdeploy/pytorch/functions/cat.py
AllentDan: I suppose it only works for `torch2onnx` with `dynamic_axies` not None.

tpoisonooo (approved):

mmdeploy/pytorch/functions/cat.py
lvhan028: @grimoire can we handle this case in the latest refactor plan?

grimoire: @lvhan028 Nope

PR #1858, opened 2023-03-08
[Enhancement] Add optional `softmax` in `LinearClsHead` 

lzhangzz:


lvhan028 (approved):

csrc/mmdeploy/codebase/mmcls/linear_cls.cpp
```diff
@@ -18,6 +18,7 @@ class LinearClsHead : public MMClassification {
  public:
   explicit LinearClsHead(const Value& cfg) : MMClassification(cfg) {
     if (cfg.contains("params")) {
+      softmax_ = cfg["params"].value("softmax", false);
```
AllentDan, line 21: `with_softmax` looks better to me. Just like `with_argmax` in segmentation.

AllentDan (approved):

csrc/mmdeploy/codebase/mmcls/linear_cls.cpp
lzhangzz: The parameter indicates the class should apply `softmax` on its input, however, `with_argmax` in segmentation indicates `argmax` is applied in the model, which has very different semantics :joy:

PR #1917, opened 2023-03-24
[Fix]: Fix mmrotate and regression script

RunningLeon:
## Motivation
Fix

## Modification

- [x] Fix oriented_rcnn+onnxruntime visualization failure. No `input_shape` is set for static deploy config.
- [x] Fix `regression_test.py`
- [x] Disable regression test for openvino and SDK+torchscript for mmpose codebase.

## BC-breaking (Optional)

None

## Use cases (Optional)

## Checklist

### DIFF
--- a/tests/regression/mmpose.yml
+++ b/tests/regression/mmpose.yml
@@ -51,7 +51,7 @@ openvino:
     deploy_config: configs/mmpose/pose-detection_openvino_static-256x192.py
   pipeline_openvino_static_fp32_256x256: &pipeline_openvino_static_fp32_256x256
     convert_image: *convert_image
-    backend_test: *default_backend_test
+    backend_test: False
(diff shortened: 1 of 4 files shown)


grimoire (approved):
LGTM

tests/regression/mmpose.yml
```diff
@@ -51,7 +51,7 @@ openvino:
     deploy_config: configs/mmpose/pose-detection_openvino_static-256x192.py
   pipeline_openvino_static_fp32_256x256: &pipeline_openvino_static_fp32_256x256
     convert_image: *convert_image
-    backend_test: *default_backend_test
+    backend_test: False
```
lvhan028, line 54: why disable the test?

RunningLeon: The performance test for openvino is too long(104125 images). The conversion result is Ok on this reg test.

PR #1921, opened 2023-03-26
fix pose_tracker python api will raise ValueError when result has no human

tongda:
## Motivation

There is a bug in Python binding of PoseTracker. When inference result has no human body, there is possibility that a `ValueError: negative dimensions are not allowed numpy` exception raised. I made some digging and found that:
1. when no human found in the image, an empty `Result` object is created but `keypoint_count` is uninitialized;
2. when creating `py::array_t` in Python version `Apply` function, `keypoint_count` is used to initialize the array shape;
3. in this case, `keypoint_count` is random value, when it is negative, the `py::array_t` would fail to create.

## Modification

Add a check before reading `keypoint_count`.

## BC-breaking (Optional)

No.

## Use cases (Optional)

No.

## Checklist


irexyc:
May change the base branch from `1.x` to `dev-1.x`

Could you please also submit the fix to master branch?

PR #1923, opened 2023-03-27
fix pose_tracker python api will raise ValueError when result has no human

tongda:
## Motivation

There is a bug in Python binding of PoseTracker. When inference result has no human body, there is possibility that a `ValueError: negative dimensions are not allowed numpy` exception raised. I made some digging and found that:
1. when no human found in the image, an empty `Result` object is created but `keypoint_count` is uninitialized;
2. when creating `py::array_t` in Python version `Apply` function, `keypoint_count` is used to initialize the array shape;
3. in this case, `keypoint_count` is random value, when it is negative, the `py::array_t` would fail to create.

## Modification

Add a check before reading `keypoint_count`.

## BC-breaking (Optional)

No.

## Use cases (Optional)

No.

## Checklist

### DIFF
--- a/csrc/mmdeploy/apis/python/pose_tracker.cpp
+++ b/csrc/mmdeploy/apis/python/pose_tracker.cpp
@@ -30,7 +30,7 @@ std::vector<py::tuple> Apply(mmdeploy::PoseTracker* self,
   std::vector<py::tuple> batch_ret;
   batch_ret.reserve(frames.size());
   for (const auto& rs : results) {
-    py::array_t<float> keypoints({static_cast<int>(rs.size()), rs[0].keypoint_count, 3});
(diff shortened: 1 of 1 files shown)


lvhan028 (approved):

alessmolt:
Hi I'm getting `ValueError: negative dimensions are not allowed numpy` even after this fix, by printing the `result`for each frame i can see that it seems to behave properly even when it detects 0 people (e.g. `(array([], shape=(0, 1819210916, 3), dtype=float32), array([], shape=(0, 4), dtype=float32), array([], dtype=uint32))`) but I'm pretty sure it is still related to this issue since when it always detect people it behaves as it is supposed until the end. Do you know what or where could be the problem?

PR #1952, opened 2023-03-31
[Enhancement]: Update dockerfile and add github release note

RunningLeon:
## Motivation

Update dockerfile and add github release note

## Modification

- Update prebuild dockerfile
- Add github release note yml

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,32 @@
+changelog:
+  categories:
+    - title: 🚀 Features
+      labels:
+        - feature
+        - enhancement
+    - title: 💥 Improvements
+      labels:
+        - improvement
+    - title: 🐞 Bug fixes
+      labels:
+        - bug
+        - Bug:P0
+        - Bug:P1
+        - Bug:P2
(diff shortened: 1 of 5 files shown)


lvhan028:
any idea how to resolve `codecov/patch`?

.github/release.yml
```diff
@@ -0,0 +1,32 @@
+changelog:
```
lvhan028, line 1: when will `release.yml` be invoked?

irexyc:
shoud update `image: openmmlab/mmdeploy:build-ubuntu18.04-cuda11.3` to `image: openmmlab/mmdeploy:build-ubuntu16.04-cuda11.3` in prebuild.yml

.github/release.yml
RunningLeon: on github release page, there is a button for triggering it manually.

irexyc (approved):

lvhan028 (approved):

PR #1977, opened 2023-04-11
Fix retinanet wrong visualization with torch1.13 

RunningLeon:
## Motivation
Fix retinanet wrong visualization with torch1.13 

## Modification
Fix retinanet wrong visualization with torch1.13 
Fix mmediting ut error
Fix typo

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/docs/en/04-supported-codebases/mmaction2.md
+++ b/docs/en/04-supported-codebases/mmaction2.md
@@ -21,7 +21,7 @@ ______________________________________________________________________
 
 ### Install mmaction2
 
-Please follow the [installation guide](https://github.com/open-mmlab/mmaction2/tree/1.x#installation) to install mmaction2.
+Please follow the [installation guide](https://github.com/open-mmlab/mmaction2/tree/main#installation) to install mmaction2.
 
 ### Install mmdeploy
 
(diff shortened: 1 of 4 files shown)


mmdeploy/codebase/mmdet/models/task_modules/coders/delta_xywh_bbox_coder.py
```diff
@@ -101,8 +101,9 @@ def delta2bbox(rois,
     dxy = denorm_deltas[..., :2]
     dwh = denorm_deltas[..., 2:]
 
-    xy1 = rois[..., None, :2]
-    xy2 = rois[..., None, 2:]
+    # fix openvino on torch1.13
+    xy1 = rois[..., :2].unsqueeze(2)
```
AllentDan, line 105: There are other places such as [this](https://github.com/open-mmlab/mmdeploy/blob/4bb9bc738c9008055fbc9347f46da70ee60fdad3/mmdeploy/codebase/mmdet/models/roi_heads/fcn_mask_head.py#L129). Shall we rewrite slice index of PyTorch?

RunningLeon: It's a model-specific problem. Have checked other modell like maskrcnn with `export_postprocess_mask=True`, and the results are OK.

Issue #1991, opened 2023-04-14
[Bug] LiteHRNet Quantization Error


moonsh:
### Checklist

- [x] I have searched related issues but cannot get the expected help.
- [x] 2. I have read the [FAQ documentation](https://github.com/open-mmlab/mmdeploy/tree/main/docs/en/faq.md) but cannot get the expected help.
- [ ] 3. The bug has not been fixed in the latest version.

### Describe the bug

The below command I tried for generating INT8 litehrnet model.

python tools/deploy.py configs/mmpose/pose-detection_ncnn-int8_static-256x192.py ../mmpose/configs/body_2d_keypoint/topdown_heatmap/coco/td-hm_litehrnet-18_8xb64-210e_coco-256x192.py ../mmpose/litehrnet18_coco_256x192.pth demo/resources/human-pose.jpg --work-dir litehrnet_ncnn --device cpu --dump-info --quant --quant-image-dir imagenet-sample-images/

However, it keeps failing when it's generating a quant table.
and error message is below

AttributeError: 'ConfigDict' object has no attribute '_env_variables'
mmengine - ERROR - tools/deploy.py - create_process - 82 - ncnn quant table failed.

### Reproduction

python tools/deploy.py configs/mmpose/pose-detection_ncnn-int8_static-256x192.py ../mmpose/configs/body_2d_keypoint/topdown_heatmap/coco/td-hm_litehrnet-18_8xb64-210e_coco-256x192.py ../mmpose/litehrnet18_coco_256x192.pth demo/resources/human-pose.jpg --work-dir litehrnet_ncnn --device cpu --dump-info --quant --quant-image-dir imagenet-sample-images/

### Environment

```Shell
04/14 08:15:09 - mmengine - INFO - **********Environmental information**********
04/14 08:15:10 - mmengine - INFO - sys.platform: linux
04/14 08:15:10 - mmengine - INFO - Python: 3.8.16 (default, Mar  2 2023, 03:21:46) [GCC 11.2.0]
04/14 08:15:10 - mmengine - INFO - CUDA available: True
04/14 08:15:10 - mmengine - INFO - numpy_random_seed: 2147483648
04/14 08:15:10 - mmengine - INFO - GPU 0: NVIDIA GeForce RTX 3090 Ti
04/14 08:15:10 - mmengine - INFO - CUDA_HOME: /usr
[...]

PR #2018, opened 2023-04-23
[Fix] Fix dataloader for ncnn quantization

irexyc:
## Motivation

fix https://github.com/open-mmlab/mmdeploy/issues/1991

## Modification

dataloader for ppq

### DIFF
--- a/tools/onnx2ncnn_quant_table.py
+++ b/tools/onnx2ncnn_quant_table.py
@@ -4,6 +4,7 @@ import logging
 from copy import deepcopy
 
 from mmengine import Config
+from torch.utils.data import DataLoader
 
 from mmdeploy.apis.utils import build_task_processor
 from mmdeploy.utils import get_root_logger, load_config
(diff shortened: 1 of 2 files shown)


RunningLeon (approved):
LGTM

RunningLeon:
@tpoisonooo @AllentDan any comment on this PR?

tools/onnx2ncnn_quant_table.py
```diff
@@ -44,16 +47,10 @@ def get_table(onnx_path: str,
...
-        if isinstance(input_tensor, list):
-            input_shape = input_tensor[0].shape
-            collate_fn = lambda x: data_preprocessor(x[0])[0].to(  # noqa: E731
-                device)
-        else:
-            input_shape = input_tensor.shape
-            collate_fn = lambda x: data_preprocessor(x)[0].to(  # noqa: E731
-                device)
-        break
+        input_tensor = input_data['inputs']
```
tpoisonooo, line 50: In some old testcase, `input_tensor` is `list`.

tpoisonooo (approved):

tools/onnx2ncnn_quant_table.py
RunningLeon: @irexyc Did we only test on mmcls models? Could we test on models from other codebases as well?

AllentDan (approved):
LGTM after resolving comments.

tools/onnx2ncnn_quant_table.py
RunningLeon: tested ok on mmpose.

PR #2027, opened 2023-04-25
[Enhancement] Load openvino model from memory

irexyc:
## Motivation

The openvino-dev version of based image is different with pip installed one, therefore the converted openvino model can't be loaded by sdk.

Previously, the openvino model will first rewrite to /tmp directory and load, which is not right when loading two openvino models at the same time. And this strategy has more overhead.

## Modification

- [x] fix cpu docker
- [x] load openvino model from memory

### DIFF
--- a/docker/CPU/Dockerfile
+++ b/docker/CPU/Dockerfile
@@ -1,4 +1,4 @@
-FROM openvino/ubuntu18_dev:2021.4.2
+FROM openvino/ubuntu20_dev:2022.3.0
 ARG PYTHON_VERSION=3.8
 ARG TORCH_VERSION=1.10.0
 ARG TORCHVISION_VERSION=0.11.0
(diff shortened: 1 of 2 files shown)


docker/CPU/Dockerfile
```diff
@@ -100,14 +100,14 @@ RUN git clone -b main https://github.com/open-mmlab/mmdeploy.git &&\
     /opt/conda/bin/mim install -e .
 
 ### build SDK
-ENV LD_LIBRARY_PATH="/root/workspace/mmdeploy/build/lib:/opt/intel/openvino/deployment_tools/ngraph/lib:/opt/intel/openvino/deployment_tools/inference_engine/lib/intel64:${LD_LIBRARY_PATH}"
+ENV LD_LIBRARY_PATH="/root/workspace/mmdeploy/build/lib:${LD_LIBRARY_PATH}"
 RUN cd mmdeploy && rm -rf build/CM* && mkdir -p build && cd build && cmake .. \
     -DMMDEPLOY_BUILD_SDK=ON \
     -DMMDEPLOY_BUILD_EXAMPLES=ON \
-    -DCMAKE_CXX_COMPILER=g++-7 \
+    -DCMAKE_CXX_COMPILER=g++-9 \
```
RunningLeon, line 107: is it necessary to upgrade to g++9?

irexyc: The base docker is ubuntu 20.04 and default gcc is 9

grimoire (approved):

RunningLeon (approved):
LGTM

PR #2029, opened 2023-04-26
[Enhanced] method of memory loading model

dhzgit:
## Motivation

There is no way to load the model from memory, so it cannot be loaded from memory when the model is encrypted

## Modification

Add methods to load models from memory

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ option(MMDEPLOY_BUILD_SDK_CSHARP_API "build SDK C# API support" OFF)
 option(MMDEPLOY_BUILD_SDK_JAVA_API "build SDK JAVA API" OFF)
 option(MMDEPLOY_BUILD_EXAMPLES "build examples" OFF)
 option(MMDEPLOY_SPDLOG_EXTERNAL "use external spdlog" OFF)
-option(MMDEPLOY_ZIP_MODEL "support SDK model in zip format" OFF)
+option(MMDEPLOY_ZIP_MODEL "support SDK model in zip format" ON)
 option(MMDEPLOY_COVERAGE "build SDK for coverage" OFF)
 option(MMDEPLOY_ELENA_FUSION "use elena to fuse preprocess" OFF)
 
--- a/csrc/mmdeploy/apis/c/mmdeploy/classifier.cpp
+++ b/csrc/mmdeploy/apis/c/mmdeploy/classifier.cpp
@@ -40,6 +40,18 @@ int mmdeploy_classifier_create_by_path(const char* model_path, const char* devic
   return ec;
 }
 
+int mmdeploy_classifier_create_by_buffer(const void* buffer, int size, const char* device_name,
+                                       int device_id, mmdeploy_classifier_t* classifier) {
+  mmdeploy_model_t model{};
+
+  if (auto ec = mmdeploy_model_create(buffer, size, &model)) {
+    return ec;
+  }
+  auto ec = mmdeploy_classifier_create(model, device_name, device_id, classifier);
+  mmdeploy_model_destroy(model);
+  return ec;
+}
+
 int mmdeploy_classifier_create_v2(mmdeploy_model_t model, mmdeploy_context_t context,
                                   mmdeploy_classifier_t* classifier) {
   return mmdeploy_pipeline_create_from_model(model, context, (mmdeploy_pipeline_t*)classifier);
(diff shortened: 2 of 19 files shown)


irexyc:
Thanks for you contribution. 

While, the current api already have the ability to load a model from memory like this:
```cpp
  mmdeploy_model_t model;
  std::ifstream ifs(model_path, std::ios::binary); // /path/to/zipmodel
  ifs.seekg(0, std::ios::end);
  auto size = ifs.tellg();
  ifs.seekg(0, std::ios::beg);
  std::string str(size, '\0');
  ifs.read(str.data(), size);
  mmdeploy_model_create(str.data(), size, &model);

  int status{};
  mmdeploy_classifier_t classifier{};
  status = mmdeploy_classifier_create(model, device_name, 0, &classifier);
  // or use mmdeploy_classifier_create_v2
```

I'm not sure if it's necessary to add these api. What is your option about this pr? @lvhan028

dhzgit:
The current code already has the usage of "mmdeploy_classifier_create_by_path", and adding the corresponding "mmdeploy_classifier_create_by_buffer" can effectively reduce misunderstandings about the API during use and corresponding usage costs. After all, even if our staff are proficient in both Python/C++and various AI model usage scenarios, it took a long time to find "mmdeploy_model_create" in the code and master the usage of this API.

CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ option(MMDEPLOY_BUILD_SDK_CSHARP_API "build SDK C# API support" OFF)
 option(MMDEPLOY_BUILD_SDK_JAVA_API "build SDK JAVA API" OFF)
 option(MMDEPLOY_BUILD_EXAMPLES "build examples" OFF)
 option(MMDEPLOY_SPDLOG_EXTERNAL "use external spdlog" OFF)
-option(MMDEPLOY_ZIP_MODEL "support SDK model in zip format" OFF)
+option(MMDEPLOY_ZIP_MODEL "support SDK model in zip format" ON)
```
lvhan028, line 35: It is recommended to keep `MMDEPLOY_ZIP_MODEL` `OFF` as the default value because not all of the users need it

csrc/mmdeploy/apis/c/mmdeploy/classifier.cpp
```diff
@@ -40,6 +40,18 @@ int mmdeploy_classifier_create_by_path(const char* model_path, const char* devic
   return ec;
 }
 
+int mmdeploy_classifier_create_by_buffer(const void* buffer, int size, const char* device_name,
```
lvhan028, line 43: We don't suggest creating another API. It's kinda redundant because `mmdeploy_model_create` and `mmdeploy_classifier_create` are enough to make up for it

lvhan028:
Hi, 

> The current code already has the usage of "mmdeploy_classifier_create_by_path", and adding the corresponding "mmdeploy_classifier_create_by_buffer" can effectively reduce misunderstandings about the API during use and corresponding usage costs. After all, even if our staff are proficient in both Python/C++and various AI model usage scenarios, it took a long time to find "mmdeploy_model_create" in the code and master the usage of this API.

I understand your concern. 
But `mmdeploy_classifier_create` already suggests reading a model from buffer, doesn't it?
And its comments say,
```
* @param[in] model an instance of mmclassification sdk model created by
 * \ref mmdeploy_model_create_by_path or \ref mmdeploy_model_create in \ref model.h
```

So, from the perspective of maintenance, I think it's not necessary to change the API. Instead, we can describe a procedure in the user guide about creating predictors from the buffer

OpenMMLab-Assistant-004:
Hi @dhzgit,

We'd like to express our appreciation for your valuable contributions to the mmdeploy. Your efforts have significantly aided in enhancing the project's quality.
It is our pleasure to invite you to join our community thorugh Discord_Special Interest Group (SIG) channel. This is a great place to share your experiences, discuss ideas, and connect with other like-minded people. To become a part of the SIG channel, send a message to the moderator, OpenMMLab, briefly introduce yourself and mention your open-source contributions in the #introductions channel. Our team will gladly facilitate your entry. We eagerly await your presence. Please follow this link to join us: ​https://discord.gg/UjgXkPWNqA.

If you're on WeChat, we'd also love for you to join our community there. Just add our assistant using the WeChat ID: openmmlabwx. When sending the friend request, remember to include the remark "mmsig + Github ID".

Thanks again for your awesome contribution, and we're excited to have you as part of our community!

lvhan028:
@irexyc could you please add a tutorial about loading mmdeploy SDK model from a buffer?

PR #2063, opened 2023-05-09
[Fix] swap width and height in Resize and Pad according to upstream codebases

AllentDan:
https://mmcv.readthedocs.io/en/2.x/understand_mmcv/data_transform.html

image

### DIFF
--- a/csrc/mmdeploy/preprocess/transform/pad.cpp
+++ b/csrc/mmdeploy/preprocess/transform/pad.cpp
@@ -83,8 +83,8 @@ class Pad : public Transform {
           data["pad_fixed_size"].push_back(pad_w);
         } else {
           padding = {0, 0, size_[0] - width, size_[1] - height};
-          data["pad_fixed_size"].push_back(size_[0]);
           data["pad_fixed_size"].push_back(size_[1]);
+          data["pad_fixed_size"].push_back(size_[0]);
         }
       } else if (size_divisor_ != 1) {
         auto pad_h = (height + size_divisor_ - 1) / size_divisor_ * size_divisor_;
--- a/csrc/mmdeploy/preprocess/transform/pad.cpp
+++ b/csrc/mmdeploy/preprocess/transform/pad.cpp
@@ -82,7 +82,7 @@ class Pad : public Transform {
           data["pad_fixed_size"].push_back(pad_h);
           data["pad_fixed_size"].push_back(pad_w);
         } else {
-          padding = {0, 0, size_[1] - width, size_[0] - height};
+          padding = {0, 0, size_[0] - width, size_[1] - height};
           data["pad_fixed_size"].push_back(size_[0]);
           data["pad_fixed_size"].push_back(size_[1]);
         }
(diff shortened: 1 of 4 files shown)


RunningLeon:
@AllentDan CI of test mmpretrain full pipeline failed.

csrc/mmdeploy/preprocess/transform/resize.cpp
```diff
@@ -74,12 +74,12 @@ class Resize : public Transform {
...
+        if (-1 == img_scale_[0]) {
           if (w < h) {
-            dst_w = img_scale_[0];
+            dst_w = img_scale_[1];
             dst_h = dst_w * h / w;
           } else {
-            dst_h = img_scale_[0];
+            dst_h = img_scale_[1];
             dst_w = dst_h * w / h;
           }
```
irexyc, line 84: When export preprocess, it will set the last dim to -1.  May not need to change this. 

https://github.com/open-mmlab/mmdeploy/blob/07e5d2d30504f884a44022ceeed5a56eb91455cb/mmdeploy/codebase/mmpretrain/deploy/classification.py#L311-L315

irexyc (changes requested):

csrc/mmdeploy/preprocess/transform/resize.cpp
AllentDan: Yes, but `img_scale_ = {height, width};` makes the first dim -1.

irexyc (approved):

RunningLeon (approved):
LGTM

csrc/mmdeploy/preprocess/transform/pad.cpp
```diff
@@ -82,9 +82,9 @@ class Pad : public Transform {
           data["pad_fixed_size"].push_back(pad_h);
           data["pad_fixed_size"].push_back(pad_w);
         } else {
-          padding = {0, 0, size_[1] - width, size_[0] - height};
-          data["pad_fixed_size"].push_back(size_[0]);
+          padding = {0, 0, size_[0] - width, size_[1] - height};
           data["pad_fixed_size"].push_back(size_[1]);
+          data["pad_fixed_size"].push_back(size_[0]);
```
irexyc, line 87: https://github.com/open-mmlab/mmcv/blob/e322848e3a466f2b1f5b2e1a9cb5552342db1e93/mmcv/transforms/processing.py#L386

pad_fixed_size seems to be in (w,h) format, although it is not used elsewhere

AllentDan: That's weird. In mmcv's document, pad_shape is formed with (height, width) which should be self.size [here](https://github.com/open-mmlab/mmcv/blob/e322848e3a466f2b1f5b2e1a9cb5552342db1e93/mmcv/transforms/processing.py#L386).

AllentDan: Check this.
> https://mmcv.readthedocs.io/en/2.x/understand_mmcv/data_transform.html
> 
> image

Issue #2144, opened 2023-06-02
[Bug] infer yolox onnx model used by python infrence sdk failed


PancakeAwesome:
used### Checklist

- [X] I have searched related issues but cannot get the expected help.
- [X] 2. I have read the [FAQ documentation](https://github.com/open-mmlab/mmdeploy/tree/main/docs/en/faq.md) but cannot get the expected help.
- [X] 3. The bug has not been fixed in the latest version.

### Describe the bug

code running until below:
bboxes, labels, _ = detector(img)
with errors:
### Reproduction

```
img = cv2.imread('mmyolo/data/6.jpg')

detector = Detector(model_path='mmdeploy_models/coco/yolox_s.py', device_name='cuda', device_id=0)
bboxes, labels, _ = detector(img)
```
errors:
Segmentation fault (core dumped)

### Environment

```Shell
### Environment

sys.platform: linux
Python: 3.8.16 (default, Mar 2 2023, 03:21:46) [GCC 11.2.0]
CUDA available: True
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 11.7, V11.7.99
GCC: gcc (GCC) 8.3.1 20190604
PyTorch: 2.0.0+cu117
TorchVision: 0.15.1+cu117
OpenCV: 4.7.0
MMEngine: 0.7.3
MMCV: 2.0.0
MMDetection: 3.0.0
MMYOLO: 0.5.0+
```
# onnx convert
python3 mmdeploy/tools/deploy.py \
    mmyolo/configs/deploy/detection_onnxruntime_static.py \
    mmyolo/yolox_s_fast_1xb32-300e-rtmdet-hyp_coco.py \
    mmyolo/yolox_s_fast_1xb32-300e-rtmdet-hyp_coco/best_coco_bbox_mAP_70.pth \
    mmyolo/data/6.jpg \
    --work-dir mmdeploy_models/coco/yolox_s_fast_1xb32-300e-rtmdet-hyp_guix \
    --device cuda:0 \
    --dump-info

used this config py: https://github.com/open-mmlab/mmyolo/blob/main/configs/deploy/detection_onnxruntime_dynamic.py

RunningLeon:
@PancakeAwesome hi,  can you run `demo/python/object_detection.py` in [here](https://github.com/open-mmlab/mmdeploy/blob/main/demo/python/object_detection.py) and post here the error log?

PancakeAwesome:
[...]

Issue #2133, opened 2023-05-30
 infer multiple images at once


oym050922021:
### Checklist

- [X] I have searched related issues but cannot get the expected help.
- [X] 2. I have read the [FAQ documentation](https://github.com/open-mmlab/mmdeploy/tree/main/docs/en/faq.md) but cannot get the expected help.
- [X] 3. The bug has not been fixed in the latest version.

### Describe the bug

How can I infer multiple images at once with the API named mmdeploy.apis.inference_model? 

### Reproduction

from mmdeploy.apis import inference_model
from glob import glob

def get_batch_image(images_list, batchsize):
      images_cnt = len(images_list)
      batch_cnt = int(images_cnt / batchsize)
      residue = images_cnt - batch_cnt * batchsize
      batch_images = []
      for i in range(batch_cnt):
            tmp = images_list[i*batchsize:(i+1)*batchsize]
            batch_images.append(tmp)
    if residue:
        tmp = images_list[-residue:]
        batch_images.append(tmp)
    return batch_images

model_cfg = "/root/Ouyangmei/openmmlab/mmdetection/configs/yolox/yolox_s_8xb8-300e_coco.py"
deploy_cfg = 'configs/mmdet/detection/detection_onnxruntime_dynamic.py'
 backend_files = ['mmdeploy_models/mmdeploy/yolox/end2end.onnx']

images_folder = '/root/Ouyangmei/openmmlab/mmdetection/datasets/teeth/images/test'
images_list = glob(images_folder + "/*.png")
 batch_images = get_batch_image(images_list, batchsize = 4)
 for images in batch_images:
        results = inference_model(model_cfg, deploy_cfg, backend_files, images, device)

### Environment

```Shell
05/30 17:45:40 - mmengine - INFO - TorchVision: 0.14.1+cu117
05/30 17:45:40 - mmengine - INFO - OpenCV: 4.7.0
05/30 17:45:40 - mmengine - INFO - MMEngine: 0.7.2
05/30 17:45:40 - mmengine - INFO - MMCV: 2.0.0rc4
05/30 17:45:40 - mmengine - INFO - MMCV Compiler: GCC 9.3
05/30 17:45:40 - mmengine - INFO - MMCV CUDA Compiler: 11.7
05/30 17:45:40 - mmengine - INFO - MMDeploy: 1.1.0+faf05fe
05/30 17:45:40 - mmengine - INFO - 

[...]

PR #2146, opened 2023-06-02
fix export info for mmyolo

RunningLeon:
## Motivation

fix https://github.com/open-mmlab/mmdeploy/issues/2144  https://github.com/open-mmlab/mmdeploy/issues/2133

## Modification

1. remove `mmdet.` domain in mmyolo pipelines
2. remove redundant score filtering before nms for yolox

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

3. The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.
4. If the modification has a dependency on downstream projects of a newer version, this PR should be tested with all supported versions of downstream projects.
5. The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/mmdeploy/codebase/mmdet/deploy/object_detection.py
+++ b/mmdeploy/codebase/mmdet/deploy/object_detection.py
@@ -268,8 +268,7 @@ class ObjectDetection(BaseTask):
         for i, transform in enumerate(transforms):
             # deal with mmyolo
             if transform['type'].startswith('mmdet.'):
-                transform['type'] = transform['type'][6:]
-                transforms[i]['type'] = transform['type']
+                transforms[i]['type'] = transform['type'][6:]
             if 'PackDetInputs' in transform['type']:
                 meta_keys += transform[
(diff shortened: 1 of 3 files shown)


mmdeploy/codebase/mmdet/deploy/object_detection.py
```diff
@@ -266,6 +266,10 @@ def get_preprocess(self, *args, **kwargs) -> Dict:
             and 'Annotation' not in item['type']
         ]
         for i, transform in enumerate(transforms):
+            # deal with mmyolo
+            if transform['type'].startswith('mmdet.'):
+                transform['type'] = transform['type'][6:]
+                transforms[i]['type'] = transform['type']
```
AllentDan, line 272: this is redundant?

RunningLeon: `transform['type']` is used later in the for loop and it should be updated as well.

AllentDan (approved):

irexyc (approved):

PR #2211, opened 2023-06-25
[Fix] Fix trt api usage error

irexyc:
## Motivation

When using trt > 8.6, there will be an error when the program ends.
```
TRTNet: 3: [runtime.cpp::~Runtime::346] Error Code 3: API Usage Error (Parameter check failed at: runtime/rt/runtime.cpp::~Runtime::346, condition: mEngineCounter.use_count() == 1. Destroying a runtime before destroying deserialized engines created by the runtime leads to undefined behavior.
```


grimoire (approved):

RunningLeon (approved):
LGTM

PR #2215, opened 2023-06-27
fix mmdeploy.version_info

xiabo123:

## Motivation

fix mmdeploy.version_info
image

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdeploy/__init__.py
+++ b/mmdeploy/__init__.py
@@ -2,7 +2,7 @@
 import importlib
 
 from mmdeploy.utils import get_root_logger
-from .version import __version__, version_info # noqa F401
+from .version import __version__, version_info  # noqa F401
 
 if importlib.util.find_spec('torch'):
     importlib.import_module('mmdeploy.pytorch')
--- a/mmdeploy/__init__.py
+++ b/mmdeploy/__init__.py
@@ -2,7 +2,7 @@
 import importlib
 
 from mmdeploy.utils import get_root_logger
-from .version import __version__  # noqa F401
+from .version import *  # noqa F401
 
 if importlib.util.find_spec('torch'):
     importlib.import_module('mmdeploy.pytorch')
--- a/mmdeploy/__init__.py
+++ b/mmdeploy/__init__.py
@@ -2,7 +2,7 @@
 import importlib
 
 from mmdeploy.utils import get_root_logger
-from .version import *  # noqa F401
+from .version import __version__, version_info # noqa F401
 
 if importlib.util.find_spec('torch'):
     importlib.import_module('mmdeploy.pytorch')


mmdeploy/__init__.py
```diff
@@ -2,7 +2,7 @@
 import importlib
 
 from mmdeploy.utils import get_root_logger
-from .version import __version__  # noqa F401
+from .version import *  # noqa F401
```
RunningLeon, line 5: hi, could you elaborate more on the motivation you use `version_info` instead of `mmdeploy.__version__`.

RunningLeon, line 5: could only import `__version__` and `version_info`

```

from .version import __version__, version_info # noqa F401
```

xiabo123: in that case, in 'version. py' “__all__ = ['__version__', 'version_info', 'parse_version_info']”,What is the meaning of it?
Function "parse_version_info", The same error occurs with info.

RunningLeon: then, we could remove `parse_version_info` from `__all__`

xiabo123: Thank you for your answer. If you don't use the feature of '__all__' in this way, this PR can be modified as you mentioned. Do you think it's okay?
`from .version import __version__, version_info # noqa F401`

xiabo123: Previously, the training referred to the "mmcv" library and compared versions using this. I encountered this problem when using 'mmdeploy' for reasoning.

RunningLeon: works for me.

RunningLeon: @xiabo123 hi, when will you update this PR?

xiabo123: Has been updated

RunningLeon:
@xiabo123 could you fix the lint error by running `pip install pre-commit && pre-commit install && pre-commit run --all-files`.

xiabo123:
ok

xiabo123:
@RunningLeon Have you encountered the following problems? During Flake8, the version number of mccabe was always incorrect. Both 0.6.0 and 0.6.1 cannot be installed in practice.
3b7e5231c69f4a9c62821ad671142cb

xiabo123:
@RunningLeon 
`from .version import __version__, version_info # noqa F401`
I'm very sorry, the modification 'version_info' was not called in 'mmdeploy' and cannot be checked by 'isort'.
I attempted to make modifications in our call layer.
`from mmdeploy.version import version_info`
We apologize for any inconvenience caused.

RunningLeon:
@xiabo123 maybe you could use here to parse package version. Cause not all packages provide `version_info`
```
import mmdeploy
from packaging import version # pip install packaging if not installed
v = version.parse(mmdeploy.__version__)
print(v.major, v.minor, v.micro)
```

xiabo123:
@RunningLeon OK,thank you.

RunningLeon (commented):
LGTM

PR #2305, opened 2023-07-27
[Fix] Prevent RuntimeError in ONNX export with empty input in ONNXNMSRotatedOp

yhna940:

## Motivation

This PR addresses a runtime error that occurs when exporting ONNX with dummy data (i.e., when no objects are present). Specifically, the issue arises from the `ONNXNMSRotatedOp` class in the `nms_rotated.py` file.

The problem was that when no objects were detected (resulting in `indices` being empty), a `RuntimeError` occurred when `torch.cat(indices)` was called. To resolve this issue, I added a condition to check the length of `indices` before concatenating. If `indices` is empty, it now returns a tensor of zeros with the appropriate shape, datatype, and device settings.

## Modification

```diff
-        indices = torch.cat(indices)
+        indices = torch.cat(indices) if len(indices) > 0 else torch.zeros(
+            (0, 3), dtype=torch.long, device=boxes.device)
```

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdeploy/mmcv/ops/nms_rotated.py
+++ b/mmdeploy/mmcv/ops/nms_rotated.py
@@ -49,7 +49,8 @@ class ONNXNMSRotatedOp(torch.autograd.Function):
                 indices.append(
                     torch.stack([batch_inds, cls_inds, box_inds], dim=-1))
 
-        indices = torch.cat(indices)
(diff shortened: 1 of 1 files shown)


RunningLeon (approved):
LGTM

AllentDan (approved):
LGTM

PR #2339, opened 2023-08-09
fix(demo): add warmup for wrong profiling result

tpoisonooo:
## Motivation

fix bad profiling result.

## Modification

add warmup

### DIFF
--- a/demo/csrc/cpp/classifier.cxx
+++ b/demo/csrc/cpp/classifier.cxx
@@ -26,10 +26,13 @@ int main(int argc, char* argv[]) {
   context.Add(mmdeploy::Device(FLAGS_device));
   context.Add(profiler);
 
+  // warmup
+  for (int i = 0; i < 20; ++i) {
+    classifier.Apply(img);
+  }
+
   // construct a classifier instance
   mmdeploy::Classifier classifier(mmdeploy::Model{ARGS_model}, context);
-
-
 
   // apply the classifier; the result is an array-like class holding references to
   // `mmdeploy_classification_t`, will be released automatically on destruction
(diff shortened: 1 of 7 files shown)


irexyc (approved):

demo/csrc/cpp/classifier.cxx
```diff
@@ -26,11 +26,14 @@ int main(int argc, char* argv[]) {
   context.Add(mmdeploy::Device(FLAGS_device));
   context.Add(profiler);
 
+  // warmup
+  for (int i = 0; i < 20; ++i) {
+    classifier.Apply(img);
+  }
+
   // construct a classifier instance
```
irexyc, line 34: should move these lines after line35

demo/csrc/cpp/rotated_detector.cxx
```diff
@@ -27,6 +27,11 @@ int main(int argc, char* argv[]) {
   context.Add(mmdeploy::Device(FLAGS_device));
   context.Add(profiler);
 
+  // warmup
+  for (int i = 0; i < 20; ++i) {
+    detector.Apply(img);
+  }
+
```
irexyc, line 34: should move these lines after line36

demo/csrc/cpp/classifier.cxx
tpoisonooo: done.

demo/csrc/cpp/rotated_detector.cxx
tpoisonooo: done

AllentDan (approved):

PR #2356, opened 2023-08-15
[Fix] Fix random color for Palette

irexyc:
The output for cv::kmeans is CV_32FC1, if doesn't reshpae c_mat, it will allocate new memory and the centers will all be zero.


RunningLeon (approved):
LGTM

PR #2392, opened 2023-08-31
fix mmdet export for coreml backend

irexyc:
## Motivation
fix mmdet export with pytorch 2.0

## Modification

add torch_alias


grimoire (approved):

RunningLeon (approved):
LGTM

PR #2426, opened 2023-09-12
fix img load for super-resolution in sdk

irexyc:
## Motivation

super-resolution task read images in rgb format 

https://github.com/open-mmlab/mmagic/blob/main/configs/srcnn/srcnn_x4k915_1xb16-1000k_div2k.py#L52-L66


RunningLeon (approved):
LGTM

AllentDan (approved):

PR #2443, opened 2023-09-18
CodeCamp2023-555

Boomerl:

## Motivation

support condinst instance segmentation algorithm from mmdet.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


mmdeploy/codebase/mmdet/models/dense_heads/condinst_head.py
```diff
@@ -0,0 +1,359 @@
...
+            points,
+            strides,
+            max_output_boxes_per_class,
+            iou_threshold,
+            score_threshold,
+            pre_top_k,
+            keep_top_k,
+            nms_type,
+        )
+
```
RunningLeon, line 115: we need to return output in `torch.Tensor` type, not InstanceData.

Boomerl: I will modify it soon.

configs/mmdet/instance-seg/instance-seg_condinst_tensorrt_dynamic-320x320-1344x1344.py
```diff
@@ -0,0 +1,21 @@
+_base_ = [
```
RunningLeon, line 1: it seems we could use this config https://github.com/open-mmlab/mmdeploy/blob/main/configs/mmdet/instance-seg/instance-seg_tensorrt_dynamic-320x320-1344x1344.py instead of adding new config

mmdeploy/codebase/mmdet/models/dense_heads/condinst_head.py
```diff
@@ -0,0 +1,359 @@
...
+        results.scores = dets[:, -1]
+        results.labels = labels
+        results.param_preds = param_preds
+        results.points = points
+        results.strides = strides
+        results_list.append(results)
+    return results_list
+
+
+def _nms_condinst(
```
RunningLeon, line 136: could we refactor and use multiclass_nms , just like rtmde-inst head
https://github.com/open-mmlab/mmdeploy/blob/01a88bee407029a2f0fea8fcb68597e6ae9bfd77/mmdeploy/codebase/mmdet/models/dense_heads/rtmdet_ins_head.py#L105

RunningLeon (changes requested):

mmdeploy/codebase/mmdet/models/dense_heads/condinst_head.py
```diff
@@ -0,0 +1,359 @@
...
+    self,
+    mask_preds: List[Tensor],
+    results_list: InstanceList,
+    batch_img_metas: List[dict],
+    rescale: bool = True,
+    **kwargs
+):
+    assert len(mask_preds) == len(results_list) == len(batch_img_metas)
+
+    for img_id in range(len(batch_img_metas)):
```
RunningLeon, line 312: we cannot for loop on batch dim, which would break batch-inference.

mmdeploy/codebase/mmdet/models/dense_heads/condinst_head.py
```diff
@@ -0,0 +1,359 @@
...
+        mask_preds.unsqueeze(0), int(self.mask_feat_stride / self.mask_out_stride)
+    ).squeeze(0)
+
+    return (mask_preds,)
+
+
+@FUNCTION_REWRITER.register_rewriter(
+    "mmdet.models.dense_heads.CondInstMaskHead.predict_by_feat"
+)
+def condinst_mask_head__predict_by_feat(
```
RunningLeon, line 302: we should remove rewritings `condinst_mask_head__forward_single` and put everything in this funciton

PR #2498, opened 2023-10-17
Fix "has no attribute startswith"

Y-T-G:
The value needs to be converted to string first.

## Motivation

When trying to 
```python
from mmdeploy.utils import get_input_shape, load_config

deploy_cfg = 'configs/mmpretrain/classification_onnxruntime_dynamic.py'
model_cfg = 'mmpretrain/configs/resnet/resnet18_8xb32_in1k.py'

# read deploy_cfg and model_cfg
deploy_cfg, model_cfg = load_config(deploy_cfg, model_cfg)
```

the `load_config` throws an error because

```python
        if model_cfg.get('_scope_', None) == 'mmrazor'\
                or model_cfg['type'].startswith('mmrazor.'):
```

 the '`type` value is not string for the function `startswith()` to work.

## Modification

Convert `type` value to string before checking the prefix.


RunningLeon (approved):
LGTM

PR #2499, opened 2023-10-17
Fix error when default_scope is None

Y-T-G:
## Motivation

Some models don't have any default scope, which means

```
        if not DefaultScope.check_instance_created(self.experiment_name):
            self.scope = DefaultScope.get_instance(
                self.experiment_name,
                scope_name=self.model_cfg.get('default_scope'))
                )
```

returns `None` which causes an error because `scope_name` should be string.

## Modification

Convert to string while retrieving value for `scope_name`


RunningLeon:
@Y-T-G hi, thanks for your PR? could you pose your env and error log when returns `None`?

RunningLeon (approved):
LGTM

PR #2506, opened 2023-10-19
Fix pose inference on list of ndarrays

Y-T-G:
## Motivation

This fixes a missing `img_path` variable error when a list of arrays are passed to `create_input()` for pose detection.

## Modification

Initialize empty list of `img_path`s for array input.


RunningLeon (approved):
LGTM

Issue #2571, opened 2023-11-28
Rtm_det-ins incorrect mask dimensions after ONNX export


LuukvandenBent:
### Checklist

- [X] I have searched related issues but cannot get the expected help.
- [X] 2. I have read the [FAQ documentation](https://github.com/open-mmlab/mmdeploy/tree/main/docs/en/faq.md) but cannot get the expected help.
- [x] 3. The bug has not been fixed in the latest version.

### Describe the bug

Rtm_det-ins incorrect mask dimensions after ONNX export. 
Related to: https://github.com/open-mmlab/mmdeploy/issues/2334

### Reproduction

 `python3 tools/deploy.py configs/mmdet/instance-seg/instance-seg_rtmdet-ins_onnxruntime_static-640x640.py mmdetection/configs/rtmdet/rtmdet-ins_s_8xb32-300e_coco.py  https://download.openmmlab.com/mmdetection/v3.0/rtmdet/cspnext_rsb_pretrain/cspnext-s_imagenet_600e.pth mmdetection/demo/demo.jpg --show --work-dir work_dir`

### Environment

```Shell
Using docker image: [ubuntu20.04-cuda11.8-mmdeploy](https://hub.docker.com/layers/openmmlab/mmdeploy/ubuntu20.04-cuda11.8-mmdeploy/images/sha256-f2b90730c5e1b2b03b775bc592b8375af1daebb1d4ff80f5bff1594073fd0524?context=explore)
Pulled latest commits and rebuild.

11/28 12:14:35 - mmengine - INFO - **********Environmental information**********
11/28 12:14:36 - mmengine - INFO - sys.platform: linux
11/28 12:14:36 - mmengine - INFO - Python: 3.8.10 (default, May 26 2023, 14:05:08) [GCC 9.4.0]
11/28 12:14:36 - mmengine - INFO - CUDA available: True
11/28 12:14:36 - mmengine - INFO - numpy_random_seed: 2147483648
11/28 12:14:36 - mmengine - INFO - GPU 0: NVIDIA GeForce RTX 3060 Laptop GPU
11/28 12:14:36 - mmengine - INFO - CUDA_HOME: /usr/local/cuda
11/28 12:14:36 - mmengine - INFO - NVCC: Cuda compilation tools, release 11.8, V11.8.89
[...]

PR #2574, opened 2023-11-29
Fix for mask dimensions mismatch in rtmdet-ins models after exporting to ONNX

LuukvandenBent:
## Motivation

This pull request is related to https://github.com/open-mmlab/mmdeploy/issues/2571.


RunningLeon (approved):
LGTM

PR #2584, opened 2023-12-01
[BUG FIX] Remove unnecessary padding during NMS 

LuukvandenBent:
## Motivation

During deployment of RtmDet-ins models, the output breaks when there are fewer predictions than the "keep_top X" setting. This is caused by padding done to the bbox but not to the masks, so when the top_k are selected, there are more bboxes than masks. This (and previous PR) is a fix for that. 

Related to https://github.com/open-mmlab/mmdeploy/issues/2571 and https://github.com/open-mmlab/mmdeploy/pull/2574.

## Modification

Removing unnecessary padding of bbox which is not done to mask.

### DIFF
--- a/mmdeploy/mmcv/ops/nms.py
+++ b/mmdeploy/mmcv/ops/nms.py
@@ -349,10 +349,6 @@ def _multiclass_nms_single(boxes: Tensor,
     dets = torch.cat([boxes, scores], dim=2)
     labels = cls_inds.unsqueeze(0)
 
-    # pad
-    dets = torch.cat((dets, dets.new_zeros((1, 1, 5))), 1)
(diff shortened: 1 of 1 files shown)


mmdeploy/mmcv/ops/nms.py
```diff
@@ -349,10 +349,6 @@ def _multiclass_nms_single(boxes: Tensor,
     dets = torch.cat([boxes, scores], dim=2)
     labels = cls_inds.unsqueeze(0)
 
-    # pad
-    dets = torch.cat((dets, dets.new_zeros((1, 1, 5))), 1)
```
RunningLeon, line 353: this padding is useful for case when `selected_indices` is zero shaped tensor.

PR #2844, opened 2024-11-19
Transfer to cpu only when torch.Tensor arguments in onnx export

take-cheeze:
## Motivation

`.cpu()` in ONNX export inputs would be only supported in Tensor inputs so it should be only applied to Tensor types

## Modification

Adds type checking before `.cpu()` and pass it directly if not.

## Checklist


lauriebax:
If this MR doesn't get merged here, feel free to open it against our VBTI fork at https://github.com/VBTI-development/onedl-mmdeploy - we're actively maintaining updated versions with latest PyTorch support and would welcome the contribution!

